Issue #33, opened 2012-07-28
Patched Font Glyphs do not appear on the Tmux Powerline.


vbajpai:
Issue: I do not see the Glyphs in the tmux powerline :-

![Imgur](http://i.imgur.com/oBZZz.png)

OS: Mac OS X 10.7.4
Terminal Application: iTerm 2

```
>> tmux -V
tmux 1.6

>> bash --version
GNU bash, version 4.2.37(2)-release (i386-apple-darwin11.2.0)

>> zsh --version
zsh 5.0.0 (x86_64-apple-darwin11.2.0)

>> ./status-left.sh 
#[bg=colour148] #[fg=colour148, bg=colour148]#[fg=colour234,bg=colour148]0:1.1 #[fg=colour148, bg=colour33]⮀#    
[fg=colour0,bg=colour33]   ilab #[fg=colour33, bg=colour24]⮀#[fg=colour255,bg=colour24]   ⓛ 10.50.203.149   #[fg=white,   
bg=colour24]⮁#[fg=colour255,bg=colour24]   ⓦ 212.201.44.246   #[fg=colour24, bg=colour235]⮀#[fg=red,bg=colour235]   #
[default]

>> ./status-right.sh
#[fg=colour237, bg=colour235]⮂#[fg=colour167,bg=colour237]    0.65 0.84 0.86 #[fg=colour37, bg=colour237]⮂#  
[fg=colour255,bg=colour37]   ✈ 26°C    #[fg=colour235, bg=colour37]⮂#[fg=colour136,bg=colour235]   Sat #[fg=default,   
bg=colour235]⮃#[fg=colour136,bg=colour235]   2012-07-28 #[fg=default, bg=colour235]⮃#[fg=colour136,bg=colour235]   14:07  
#[default]
```

My tmux configuration:

```
# split panes
bind | split-window -h
bind - split-window -v

# resize panes
bind -r h resize-pane -L 5
bind -r j resize-pane -D 5
bind -r k resize-pane -U 5
bind -r l resize-pane -R 5

# enable 256-colors
set -g default-terminal "screen-256color"

# enable utf-8 on status bar
set -g status on
set -g status-utf8 on

# status refresh interval
set -g status-interval 2

# status left
set -g status-left-length 60
set -g status-left "#(~/.tmux/tmux-powerline/status-left.sh)"

# status right
set -g status-right-length 90
set -g status-right "#(~/.tmux/tmux-powerline/status-right.sh)"

# solarized theme
[...]

Issue #190, opened 2014-10-31
powerline fonts doesn't work


christian-bromann:
Hi,

I am running Mac OS X Yosemite and tried a lot of stuff to get the fonts imported but for some reasons I get this:

![screen shot 2014-10-31 at 15 49 36](https://cloud.githubusercontent.com/assets/731337/4862500/2c580f86-610d-11e4-9048-71811918c7b5.png)

Any idea why this happens?

yukie-:
hey everyone, I hope you ended up figuring this out. This is due to not having patched fonts. make sure you're using this repo https://github.com/powerline/fonts

![fontsss](https://cloud.githubusercontent.com/assets/10278096/8994521/e0c837fa-36d9-11e5-8e0c-eb06fa5ca940.png)

I'm still configuring my .tmux.conf but you can see here the fonts are at least patched correctly. 

I'm on Yosemite 10.10.13 

Hope this helps!

erikw:
I'm closing this stale issue due to age.

PR #193, opened 2015-02-15
fixed separator characters to work with powerline fonts

MoOmEeN:
fix for issue reported [here] (https://github.com/erikw/tmux-powerline/issues/33#issuecomment-14959749)


MoOmEeN:
fix for https://github.com/erikw/tmux-powerline/issues/190

erikw:
Well it was never broken since tmux-powerline uses glyphs from the older versions of the powerline patched fonts. But I'm OK with bringing in this patch as I guess the new patched fonts are more common nowadays :)

MoOmEeN:
Hi,
Yeah. I realised that it uses different font bit too late. But imho good
idea to switch to this font since vim-airline and oh-my-zsh themes use it
already.

Thanks
MoOmEeN


On Pn, lut 16, 2015 at 19:41 , Erik Westrup <notifications@github.com>
wrote:

Well it was never broken since tmux-powerline uses glyphs from the older
versions of the powerline patched fonts. But I'm OK with bringing in this
patch as I guess the new patched fonts are more common nowadays :)


—
Reply to this email directly or view it on GitHub
<https://github.com/erikw/tmux-powerline/pull/193#issuecomment-74554029>.










-- 
Sent from MetroMail

MoOmEeN:
@erikw looks like you merged also my another commit. That wasn't my idea, that other one is just my fix for gmail count feature. Might be useful as well but probably as separate commit slightly modified (using grep instead of ggrep)

erikw:
Yup I saw that, so I reverted that one directly afterwards :)

-- 
Erik Westrup
<erik.westrup@gmail.com>
(+46) 0738-286060 | 2r.se
Lagerbrings väg 8E, 22460 Lund, Sweden
-----------------------------------------
This email is encrypted with 2ROT-13.

PR #195, opened 2015-02-20
Fix mailcount xml fullcount tag parsing

mullakhmetov:
Added OS X Yosemite 10.10.2 gmail mailcount compatibility: fixed xml fullcount-tag parsing


erikw:
I removed the if shell_osx because the old grep did not work any more for Linux it seems. My theory is that before, the output from the query was on multiple lines but now it's only one line.

PR #196, opened 2015-02-22
added OS X compability in ifstat_sys

mullakhmetov:
There is no /sys/ in OS X, so i get I/O bytes info on en0 via netstat.


erikw:
Cool

PR #197, opened 2015-05-01
Support OSX/BSD grep and date

echiu64:
set TMUX_POWERLINE_SEG_WEATHER_GREP to egrep for OSX/BSD

Tested on OSX only


echiu64:
Refactored date code per suggestions

erikw:
:clap:

PR #201, opened 2015-08-01
Corrected echo string, removed duplicate "must be"

yukie-:


erikw:
Thanks! =)

yukie-:
happy I could help  :)

PR #207, opened 2015-11-24
Added the option to control RCFILE trough env

DxCx:
Hi,

I need the ability to configure the path for the RC file,
you are welcome to merge if you like :)


erikw:
Hmm I suggest making the default rc file depend on that too, so

export TMUX_POWERLINE_RCFILE_DEFAULT="${TMUX_POWERLINE_RCFILE}.default"

DxCx:
There you go, fixed as suggest,
i agree it should be that way =)

erikw:
Thanks!

PR #210, opened 2016-01-14
fix newline in file

Kshitij-Banerjee:
New line character in file content, makes things awry and eats up the time and weather segments. This minor fix will make the contents of the file come into a single file with a | delimiter.


dbrgn:
For the record, this is to add support for the [orochi](https://github.com/dbrgn/orochi) format that logs the current song information (track, artist, album, year) separated by newline charcters.

PR #212, opened 2016-02-11
show error when rainbarf is not found

long-long-float:
I installed `rainbarf` via Linuxbrew and enabled rainbarf segment, but rainbarf segment was not shown. It took time to find the cause of this phenomenon.

I modified to show error when rainbarf is not found, and foreground color because this error can be read.


erikw:
Thanks!

PR #215, opened 2016-03-09
no need to set TMUXPWD

chaoswork:
use pane_current_path, no need to set TMUXPWD
tested in tmux2.1 and tmux1.9a

### DIFF
--- a/lib/tmux_adapter.sh
+++ b/lib/tmux_adapter.sh
@@ -1,8 +1,8 @@
 # Get the current path in the segment.
 get_tmux_cwd() {
 	local env_name=$(tmux display -p "#D" | tr -d %)
-	#local env_val=$(tmux show-environment | grep --color=never "$env_name")
-  local env_val=$(tmux display-message -p -F "#{pane_current_path}" -t"$env_name")
+	# use papne_current_path, no need to set TMUXPWD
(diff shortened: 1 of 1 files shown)


lib/tmux_adapter.sh
```diff
@@ -1,7 +1,8 @@
 # Get the current path in the segment.
 get_tmux_cwd() {
-	local env_name=$(tmux display -p "TMUXPWD_#D" | tr -d %)
-	local env_val=$(tmux show-environment | grep --color=never "$env_name")
+	local env_name=$(tmux display -p "#D" | tr -d %)
+	# use pane_current_path, no need to set TMUXPWD
+	local env_val=$(tmux display-message -p -F "#{pane_current_path}" -t"$env_name")
```
erikw: I tried this out in a tmux session, and it seems like -t to display-message do want a % before the name, otherwise I get e.g. "can't find pane 15".

If I remove the "tr -d %" it works for me. I'm using Tmux v2.1.

erikw:
See my comment on the patch. Also please squash your three commits before making pull requests. Some development history is good to have, but the detail in these small commits are not relevant I think :)

nomaed:
My `get_tmux_cwd` command is just this single line:
```bash
get_tmux_cwd() {
    tmux display-message -p "#{pane_current_path}"
}
```

If the goal is to print the PWD of the currently selected pane, this does the job.

PR #218, opened 2016-03-28
Allow the music player to be a list

duckpuppy:
The `$TMUX_POWERLINE_SEG_NOW_PLAYING_MUSIC_PLAYER` variable can be
specified as a comma-separated list. Each item in the list will be
checked in order, and the first one that is found playing will be used
to construct the segment.

Example: "spotify,itunes,mpd"


erikw:
Looks cool!

PR #222, opened 2016-04-26
Enhanced spotify applescript

sherifkandeel:
it now shows a little play/pause icon and the song information still shows when paused.

### DIFF
--- a/segments/np_spotify_mac.script
+++ b/segments/np_spotify_mac.script
@@ -7,12 +7,17 @@ end tell
 
 if process_list contains "Spotify" then
 	tell application "Spotify"
-		if player state is playing then
+		# if player state is playing then
+		if player state is playing or player state is paused then
 			set track_name to name of current track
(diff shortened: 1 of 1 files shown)


segments/np_spotify_mac.script
```diff
@@ -7,17 +7,23 @@ end tell
...
+				set now_playing to "❙❙ " & artist_name & " - " & track_name
+			end if
 			if length of now_playing is less than trim_length then
 				set now_playing_trim to now_playing
 			else
 				set now_playing_trim to characters 1 thru trim_length of now_playing as string
 			end if
 		end if
 	end tell
+else
```
erikw: I think that if the music is stopped, then the segment should be hidden (no output here). That's how the other NP segments work.

sherifkandeel: @erikw you have a point, I fixed that.

erikw:
Thanks!

PR #223, opened 2016-05-02
replaced battery equation with a more accurate osx command

sherifkandeel:
Apparently getting the battery current capacity and dividing that over the max capacity gives a wrong estimate about the battery remaining percentage, at least from Mac OS point of view.
(I am assuming they have a statistical model to base the remaining time on, factoring in the user behaviour, the open apps...etc)

I found that using `pmset` gives exactly the percentage that appears on top screen. so I decided that is a more accurate measure, instead of the annoyance of having two different values.

I left the other command with its variables intact, in order not to break the current behaviour.

### DIFF
--- a/segments/battery.sh
+++ b/segments/battery.sh
@@ -66,7 +66,7 @@ __battery_osx() {
 				if [[ "$curcap" == "$maxcap" || "$fully_charged" == "Yes" && $extconnect == "Yes"  ]]; then
 					return
 				fi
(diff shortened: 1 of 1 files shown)


erikw:
I don't have an Mac to test with anymore, but looks cool!

sherifkandeel:
I have tested this on my main Mac. and the old method is off by 5~15% depending on your running apps.
Thanks for merging.

PR #224, opened 2016-06-14
updated uptime.sh with support for alternate 'grep' on Mac OS X

ankurs:
uptime.sh currently doesn't support using a different 'grep' command, this breaks in Mac OS X as default grep is much older version

i have added support for switching default grep so that uptime.sh works in Mac OS X

### DIFF
--- a/segments/uptime.sh
+++ b/segments/uptime.sh
@@ -1,6 +1,29 @@
 # Prints the uptime.
 
+
+if shell_is_bsd; then
+    TMUX_POWERLINE_SEG_UPTIME_GREP_DEFAULT="/usr/local/bin/grep"
+else
+    TMUX_POWERLINE_SEG_UPTIME_GREP_DEFAULT="grep"
+fi
+
+__process_settings() {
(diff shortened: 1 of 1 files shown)


segments/uptime.sh
```diff
@@ -1,6 +1,29 @@
 # Prints the uptime.
 
+
+if shell_is_bsd; then
+    TMUX_POWERLINE_SEG_UPTIME_GREP_DEFAULT="/usr/local/bin/grep"
```
erikw: Is grep usually installed in /usr/local for BSDs?

ankurs: From what i have seen, we are using the same default in 'weather.sh' as well.

erikw: Fair enough :)

erikw:
Thanks

segments/uptime.sh
Meldanya: @erikw I think he's assuming that a newer version of (GNU) grep is installed in /usr/local by brew. My grep is in /usr/bin.

PR #225, opened 2016-06-23
Added PWD update command for Fish shell

nomaed:


erikw:
Danke schön :)

nomaed:
Bitte :)
However, looking at the PR from @chaoswork, this step can be easily removed. I'm testing it myself now.

PR #229, opened 2016-07-14
Fix for wrong VCS information

### LINKED ISSUE #228: Git VCS segments sometimes show wrong information
Hey,
thanks for the great project! I have worked with tmux powerline for some days now and noticed that the vcs segments sometimes show wrong information for git repositories (or I don't understand the segments correctly). What I noticed:

- `vcs_modified` shows the count of modified files regardless of their staging status. I think it would be better to exclude staged files because these are already recorded by `vcs_staged`.
- `vcs_others` and `vcs_modified` only count untracked and unstaged files starting from the current working directory, whereas `vcs_staged` considers the complete git repository.
- `vcs_compare` shows odd numbers if git is in detached head mode.

I've written a small patch that modifies the vcs segment behavior in the following way:

- `vcs_modfied` only considers unstaged changes.
- Untracked, modified and staged files are always counted in the whole git repository regardless of the current working directory.
- `vcs_compare` is deactivated if git is in detached head mode since there is no upstream reference.

IngoHeimbach:
This patch fixes #228

### DIFF
--- a/segments/vcs_compare.sh
+++ b/segments/vcs_compare.sh
@@ -33,7 +33,10 @@ __parse_git_stats() {
 	# check if git
 	[[ -z $(git rev-parse --git-dir 2> /dev/null) ]] && return
 
-	tracking_branch=$(git for-each-ref --format='%(upstream:short)' $(git symbolic-ref -q HEAD))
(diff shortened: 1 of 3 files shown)


erikw:
Cool thanks! :)

PR #232, opened 2016-07-30
Segments uptime bugfix

flytreeleft:


flytreeleft:
Contains useless commits, should be deleted!

Issue #235, opened 2016-08-07
dbus call broken for spotify


IngoHeimbach:
Hi,
the dbus api for spotify has been changed, so the "now playing" segment is currently broken for spotify on Linux machines. Just change [this line](https://github.com/erikw/tmux-powerline/blob/master/segments/now_playing.sh#L273):

``` bash
metadata=$(dbus-send --reply-timeout=42 --print-reply --dest=org.mpris.MediaPlayer2.spotify / org.freedesktop.MediaPlayer2.GetMetadata 2>/dev/null)
```

to

``` bash
metadata=$(dbus-send --reply-timeout=42 --print-reply --session --dest=org.mpris.MediaPlayer2.spotify /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties.Get string:'org.mpris.MediaPlayer2.Player' string:'Metadata' 2>/dev/null)
```

and spotify integration works again on linux. :)

PR #236, opened 2016-08-07
Spotify's dbus api has been changed -> fixed the dbus call

IngoHeimbach:
Fix for #235


erikw:
That is true, verified. Thanks!!

PR #237, opened 2016-08-14
Enhancement: Ability to use RGB colors in themes

nomaed:
This patch adds the ability to use RGB colors in the #RRGGBB format alongside the classic 0-255 color indices.

### DIFF
--- a/lib/formatting.sh
+++ b/lib/formatting.sh
@@ -1,7 +1,7 @@
 __print_colored_content() {
-    [ "${2:0:1}" = "#" ] && BGCOLOR="$2" || BGCOLOR="colour$2"
-    [ "${3:0:1}" = "#" ] && FGCOLOR="$3" || FGCOLOR="colour$3"
-    echo -n "#[fg=${FGCOLOR},bg=${BGCOLOR}]"
-    echo -n "$1"
-    echo -n "#[default]"
+	[ "${2:0:1}" = "#" ] && BGCOLOR="$2" || BGCOLOR="colour$2"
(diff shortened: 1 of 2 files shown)


lib/formatting.sh
```diff
@@ -1,5 +1,7 @@
 __print_colored_content() {
-	echo -n "#[fg=colour$3, bg=colour$2]"
-	echo -n "$1"
-	echo -n "#[default]"
+    [ "${2:0:1}" = "#" ] && BGCOLOR="$2" || BGCOLOR="colour$2"
```
nomaed: You might be right. I think my default 2-spaces indent got in. Fixing...

lib/powerline.sh
```diff
@@ -101,7 +101,7 @@ __process_powerline() {
 		local separator_background_color=${powerline_segment[4]}
 		local separator_foreground_color=${powerline_segment[5]}
 
-		eval "__print_${side}_segment ${segment_index} ${background_color} ${foreground_color} ${separator} ${separator_background_color} ${separator_foreground_color}"
+		eval "__print_${side}_segment \"${segment_index}\" \"${background_color}\" \"${foreground_color}\" \"${separator}\" \"${separator_background_color}\" \"${separator_foreground_color}\""
```
erikw: Why did you add escaped doublequotes? :)

nomaed: Because `eval`.
Are you asking why it's double quotes as opposed to single quotes?
Because in general, each segment has to be passed to `__print_*_segment` as a single argument, so it needs to be enclosed in quotes.

PR #239, opened 2016-09-30
Modify don't create index.lock on vcs_modified and vcs_others.sh

naoto-k:
Recently, I encount lot of index.lock on my local git repository. To the best of my knowledge, git-status command on vcs_xxx.sh creates lock file. So I modified to use ls-files instead of git-status.


erikw:
Cool thanks!!

PR #240, opened 2016-10-29
Change to display without omitting the branch name

TakkuMattsu:
Branch name is displayed as "XXXXX" when there is a branch called "topic / XXXXX".
So I want to display without omitting the branch name.

## result
image


erikw:
Geil! :)

PR #241, opened 2016-11-05
no logger necessary set-option -g status-utf8 on

mingchaoyan:


erikw:
It's default nowadays you mean??

mingchaoyan:
https://github.com/tmux/tmux/issues/230

erikw:
Thx!

PR #245, opened 2017-01-31
New segment - prefix_mode_indicator

dominikduda:
Simple segment to indicate whenever TMUX prefix is active.


erikw:
Cool! Is it possible to detect copy-mode as well?

dominikduda:
I will check whenever I will have some free time

erikw:
Anyways this works, I'll merge. Thanks!

PR #247, opened 2017-02-02
Update prefix_mode_indicator

dominikduda:
I added copy mode indicator and updated look.

Right now it should look like this:
image
image
image
image


dominikduda:
I also think that it could be added to default  theme as it has no dependencies and is pretty straightforwart and usefull. What do you think?

erikw:
Looks dank man, good job! :)

PR #248, opened 2017-02-08
Added a segment for querying osx notification counts

IngoHeimbach:
The segment reads an internal sqlite database of Apple's notification
center to get the count of notifications for one or multiple apps
(Apple does not provide an apple script interface for that).

This segment is especially useful for getting the count of unread messages in Messages.app.

This approach is a bit hacky since internal data structeres are read but I don't know how it could be done better. :(

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,3 +32,4 @@ antiAgainst <antiAgainst@gmail.com>
 Alexander Luberg <alex@luberg.me>
 Stanislaw Pusep <stas@sysd.org>
 Austin Beam
+Ingo Heimbach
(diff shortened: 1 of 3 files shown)


segments/osx_notification_count.sh
```diff
@@ -0,0 +1,52 @@
+# Report mail count for your mailbox type
```
erikw: Notification count *

IngoHeimbach:
Thanks for your review, the comment should be ok now. :)

erikw:
So I tried it out now, works perfect! :)

Add yourself to the `AUTHORS` file if you want and consider renaming OSX to macOS as that is the new and current name for this OS :)

IngoHeimbach:
Done! :)

erikw:
:D

PR #249, opened 2017-02-16
Renamed all occurrences of "OSX" to "MACOS" in the macos notification count plugin

IngoHeimbach:
Sorry, I have forgotten to rename the shell variables from "OSX" to "MACOS"... :(


erikw:
:)

PR #250, opened 2017-04-29
fixed incorrent batt percentage

yashwanthreddyg:
Fixed issue where battery percentage is greater than 100 when charge_now is greater than charge_full.


erikw:
Thank you!

PR #251, opened 2017-05-10
location not set bugfix

fr1sk:


segments/weather.sh
```diff
@@ -1,20 +1,15 @@
-# Prints the current weather in Celsius, Fahrenheits or lord Kelvins. The forecast is cached and updated with a period of $update_period.
```
erikw: Why remove this comment?

segments/weather.sh
```diff
@@ -1,20 +1,15 @@
...
 
 # The update period in seconds.
 update_period=600
 
 TMUX_POWERLINE_SEG_WEATHER_DATA_PROVIDER_DEFAULT="yahoo"
 TMUX_POWERLINE_SEG_WEATHER_UNIT_DEFAULT="c"
 TMUX_POWERLINE_SEG_WEATHER_UPDATE_PERIOD_DEFAULT="600"
-if shell_is_bsd; then
```
erikw: Why remove BSD grep check?

segments/weather.sh
```diff
@@ -1,20 +1,15 @@
...
-	TMUX_POWERLINE_SEG_WEATHER_GREP_DEFAULT="/usr/local/bin/grep"
-else
-	TMUX_POWERLINE_SEG_WEATHER_GREP_DEFAULT="grep"
-fi
-
+TMUX_POWERLINE_SEG_WEATHER_GREP_DEFAULT="grep"
+TMUX_POWERLINE_SEG_WEATHER_LOCATION=""
 
 generate_segmentrc() {
-	read -d '' rccontents  << EORC
```
erikw: The original indent was intended! :) Contents inside the function is indented, but not the heredoc :)

segments/weather.sh
```diff
@@ -28,9 +23,9 @@ export TMUX_POWERLINE_SEG_WEATHER_GREP="${TMUX_POWERLINE_SEG_WEATHER_GREP_DEFAUL
 # 1. Go to Yahoo weather http://weather.yahoo.com/
 # 2. Find the weather for you location
 # 3. Copy the last numbers in that URL. e.g. "http://weather.yahoo.com/united-states/california/newport-beach-12796587/" has the numbers "12796587"
-export TMUX_POWERLINE_SEG_WEATHER_LOCATION=""
+export TMUX_POWERLINE_SEG_WEATHER_LOCATION="${TMUX_POWERLINE_SEG_WEATHER_LOCATION_DEFAULT}"
```
erikw: This default value is not defined among the other default values, starting at line 6 above.

What would a good default value for a location be? I don't know, that's why there is not default set.

The intended usage here is simply that the user sets the value of `TMUX_POWERLINE_SEG_WEATHER_LOCATION` in `/.tmux-powerlinerc` to something.

:-)

PR #252, opened 2017-05-22
Use tmux to query the cwd of the active pane

### LINKED ISSUE #188: manipulation of PS1 is not needed
Tmux provides `#{pane_current_path}` which can be read in external scripts:

`PANE_PWD="$( tmux display -p '#{pane_current_path}' )"`

IngoHeimbach:
Since version 2.1 `tmux display` can be used to query the current
working directory in the active pane. This commit makes use of this
query function if tmux is recent enough. Otherwise, the old query code
is used as a fallback.

Fixes #188

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,8 @@ bind C-[ run '~/path/to/tmux-powerline/mute_powerline.sh left'		# Mute left stat
 bind C-] run '~/path/to/tmux-powerline/mute_powerline.sh right'		# Mute right statusbar.
 ```
 
+## For tmux versions < 2.1
+
 Some segments e.g. cwd and cvs_branch needs to find the current working directory of the active pane. To achieve this we let tmux save the path each time the shell prompt is displayed. Put the line below in your `~/.bashrc` or where you define you PS1 variable. zsh users can put it in e.g. `~/.zshrc` and may change `PS1` to `PROMPT` (but that's not necessary).
 
 ```bash
@@ -137,6 +139,8 @@ If the active shell is Fish, PS1 is not being set normally. Instead, it has a `f
 ```fish
 if set -q TMUX; tmux setenv TMUXPWD_(tmux display -p "#D" | tr -d '%') $PWD; end
 ```
+
+If you have a recent version of tmux (&ge; 2.1), there is no need to redefine the PS1 variable since tmux can be called directly to query the working directory of the active pane.
 
 # Configuration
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -138,6 +140,8 @@ If the active shell is Fish, PS1 is not being set normally. Instead, it has a `f
 if set -q TMUX; tmux setenv TMUXPWD_(tmux display -p "#D" | tr -d '%') $PWD; end
 ```
 
+If you have a recent version of tmux (> 2.0), there is no need to redefine the PS1 variable since tmux can be called directly to query the working directory of the active pane.
```
erikw: `> 2.1`?

lib/tmux_adapter.sh
```diff
@@ -1,12 +1,32 @@
...
-	local env_val=$(tmux show-environment | grep --color=never "$env_name")
-	# The version below is still quite new for tmux. Uncomment this in the future :-)
-	#local env_val=$(tmux show-environment "$env_name" 2>&1)
-
-	if [[ ! $env_val =~ "unknown variable" ]]; then
-		local tmux_pwd=$(echo "$env_val" | sed 's/^.*=//')
-		echo "$tmux_pwd"
+MIN_MAJOR_VERSION="2"
+MIN_MINOR_VERSION="1"
+TMUX_VERSION="$(tmux -V)"
```
erikw: I know `tmux -V` is quite not, I think it was not added until `v1.9` or so, but maybe the check at line 6 handles this anyways, simply by not matching the patterns?  :)

README.md
IngoHeimbach: `pane_current_path` is available since version 2.1, so maybe `>= 2.1`?

lib/tmux_adapter.sh
IngoHeimbach: I've searched the commti history with `git grep` and found out that `-V` was already present in v0.1 from November 2017 (see commit [0b832ea4](https://github.com/tmux/tmux/commit/0b832ea466cb013de2530f230d9a9ea9f1cd5641)). Or has the switch been removed in other versions?

IngoHeimbach:
I simply amended the old commit and sent the new one by force push - I hope that was not wrong?

lib/tmux_adapter.sh
erikw: Hmm looks so from the commit yes. I just remember that not so long ago `-V` did not work on the tmux versions shipped on Ubuntus for example. Lets try this and see if someone complains ;)

PR #258, opened 2018-03-23
Fix weather segment on OSX

msempere:
**Following PR:**
 - Stops using `/usr/local/bin/grep` if path is not existent for OSX (maybe previously was installed using brew, and this puts the stuff under `/usr/local/`). Current OSX grep works fine, and is no longer part of `/usr/local`. But keeping the previous path for compatibility.

 - Moves `TMUX_POWERLINE_SEG_WEATHER_LOCATION` where it can be used. Previously that var was not being populated following the instructions.

**Output proof:**

screen shot 2018-03-23 at 23 20 25


erikw:
Cool. I will take a closer look when I'm at home

Feel free to append yourself to https://github.com/erikw/tmux-powerline/blob/master/AUTHORS in the PR

msempere:
Tested on Linux and worked fine as well. Let me know what you think.

PR #259, opened 2018-03-24
Update README.md with new brew command

msempere:
Following PR:
 - Updates README.md with new instructions to upgrade `grep` on OS X 10.8 Mountain Lion. `homebrew/dupes` is deprecated and has been moved to `homebrew/core`.


erikw:
(y) thx!

PR #260, opened 2018-03-25
Fix uptime segment

msempere:
**Following PR:**
 - Moves logic from `grep` to `sed`
 - Fixes issue where uptime is not working on OSX

**Proof execution:**

screen shot 2018-03-25 at 17 54 27


erikw:
LGTM

PR #264, opened 2018-04-24
fix ifstat_sys.sh to support systemd

frostnotfall:
In some systemd-based system，the network interface is no longer eth0, but is named like ensXX。
In those system，"ifstat_sys.sh" may not work properly.
This change will automatically find the correct name of the network interface.


erikw:
Nifty AWK :), looks like it works.

PR #277, opened 2021-12-17
Show current keyboard layout for Gnome40

SemyonDunaev:
Use Gnome40 gsettings to get and show current used keyboard layout.

### DIFF
--- a/segments/gnome4_layout.sh
+++ b/segments/gnome4_layout.sh
@@ -1,0 +1,13 @@
+# Print the currently used keyboard layout
+# Works with Gnome 40
+# Set smaller update interval in .tmux.conf: 'set-option -g status-interval 1'
+# Exit if platform is not linux as this script is dependant on Linux and Gnome DE
+
+run_segment() {
+	if ! shell_is_linux; then
+		return 1
+	fi
+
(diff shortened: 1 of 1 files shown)


erikw:
Hey thanks for your contribution!

Just one comment: let's return a proper exit value also in the case we're on a Linux system. The reurn value is used here: https://github.com/erikw/tmux-powerline/blob/9ee6bc491088606390c75ac193e78ab6597302af/lib/powerline.sh#L49-L50

Probably capture the reurn value of the gsettings line and rerun with this

```bash
cur_layout=$(gsettings get org.gnome.deskt...) 
ecode=$? 
test $ecode  -0 eq || return $ecode

echo "⌨  $cur_layout"
return 0
```

Something like this :)

SemyonDunaev:
> Probably capture the reurn value of the gsettings line and rerun with this
> 
> ```shell
> cur_layout=$(gsettings get org.gnome.deskt...) 
> ecode=$? 
> test $ecode -eq 0 || return $ecode
> 
> echo "⌨  $cur_layout"
> return 0
> ```
> 
> Something like this :)

Sory, i'm not very good at this.ill try )

erikw:
@SemyonDunaev thank you for your contribution. And happy new year! :)

PR #278, opened 2022-01-04
Adding tun0 ip address to the segmants bar when the ip address is available.

CrunchV1:
It's quite nice when working with for example hackthebox or tryhackme.

### DIFF
--- a/themes/default.sh
+++ b/themes/default.sh
@@ -70,6 +70,7 @@ if [ -z $TMUX_POWERLINE_RIGHT_STATUS_SEGMENTS ]; then
 	TMUX_POWERLINE_RIGHT_STATUS_SEGMENTS=(
 		#"earthquake 3 0" \
 		"pwd 89 211" \
+		"tun0 13 255" \
 		#"macos_notification_count 29 255" \
 		"mailcount 9 255" \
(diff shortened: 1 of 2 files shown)


themes/default.sh
```diff
@@ -70,6 +70,7 @@ if [ -z $TMUX_POWERLINE_RIGHT_STATUS_SEGMENTS ]; then
 	TMUX_POWERLINE_RIGHT_STATUS_SEGMENTS=(
 		#"earthquake 3 0" \
 		"pwd 89 211" \
+		"tun0 13 255" \
```
erikw, line 73: As it's not common to have a tun interface, let's not add this to the default theme. It's easy to make your own theme outside that includes it :)

erikw (changes requested):
Thanks for the PR. Just one thing to fix :)

themes/default.sh
CrunchV1: Gotcha! Thanks for you time!

erikw:
@CrunchV1 let's get this PR closed! It's just a small change to fix :)

erikw:
@CrunchV1 Ah sorry I meant closed as it merged!!

PR #283, opened 2022-03-02
Add support for `playerctl` in the `now_playing` segment

IngoMeyer441:


segments/now_playing.sh
```diff
@@ -121,6 +125,9 @@ __process_settings() {
 	if [ -z "$TMUX_POWERLINE_SEG_NOW_PLAYING_NOTE_CHAR" ]; then
 		export TMUX_POWERLINE_SEG_NOW_PLAYING_NOTE_CHAR="${TMUX_POWERLINE_SEG_NOW_PLAYING_NOTE_CHAR_DEFAULT}"
 	fi
+	if [ -z "$TMUX_POWERLINE_SEG_NOW_PLAYING_PLAYERCTL_FORMAT" ]; then
+		export TMUX_POWERLINE_SEG_NOW_PLAYING_PLAYERCTL_FORMAT="${TMUX_POWERLINE_SEG_NOW_PLAYING_PLAYERCTL_FORMAT}"
```
erikw, line 129: The value should be `TMUX_POWERLINE_SEG_NOW_PLAYING_PLAYERCTL_FORMAT_DEFAULT` right?

I think it's the same issue on `TMUX_POWERLINE_SEG_NOW_PLAYING_RHYTHMBOX_FORMAT` below as well. Maybe you can fix it too? :)

erikw (changes requested):

erikw:
Thank you @IngoMeyer441. This PR looks good, just one comment to address :)

segments/now_playing.sh
IngoMeyer441: Ah, yes this is a copy-and-paste mistake :innocent: 
Both is fixed in the latest commit.

erikw:
Thank you!

PR #285, opened 2022-05-03
Switched out deprecated yahoo API with met.no

Mraedis:
Hello, I've implemented met.no as a weather provider.
I removed the yahoo one outright because it is no longer accessible, and commented out the sunrise/sunset part because this is included in the results met.no gives.

Feel free to use it, change it or disregard it.


erikw:
Really nice @Mraedis. I tried it out and it worked great! Thank you


Screen Shot 2022-05-04 at 12 15 05

Solves @219

PR #286, opened 2022-06-04
Add utc time formatting

kurczynski:
Add the ability to format the utc_time segment.


erikw:
I tried it out, works fine. Thanks!

Issue #298, opened 2022-12-02
Allow tmux session info to be customized


tnkhanh:
In `segments/tmux_session_info.sh` , the format is fixed to `'#S:#I.#P'` : [ tmux_session_info.sh](https://github.com/erikw/tmux-powerline/blob/aaef83e1c4a6616a067eeac8fb3d483d053808a2/segments/tmux_session_info.sh#L1-L7)

Although powerline user can change the format in this file, it's better to have the format modifiable in `.tmux-powerlinerc` (I think).

For example, `date.sh` has the format modifiable in `export TMUX_POWERLINE_SEG_DATE_FORMAT="%d %b %+4Y"` in the rc file. Maybe we can do something similar to `tmux_session_info.sh`

(Or we can change `date.sh` to be modifiable at source like `tmux_session_info.sh`, it's weird they are different)

erikw:
A PR is welcomed, there are plenty of examples in https://github.com/erikw/tmux-powerline/tree/main/segments how this can be done :)

tnkhanh:
I submitted a PR to resolve this issue

erikw:
Fixed by https://github.com/erikw/tmux-powerline/pull/299

PR #299, opened 2022-12-04
Allow session info format to be modifiable in rc file

tnkhanh:
This PR solves https://github.com/erikw/tmux-powerline/issues/298

### DIFF
--- a/segments/tmux_session_info.sh
+++ b/segments/tmux_session_info.sh
@@ -1,7 +1,24 @@
 # Prints tmux session info.
-# Assuems that [ -n "$TMUX"].
+# Assumes that [ -n "$TMUX"].
+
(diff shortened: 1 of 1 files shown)


segments/tmux_session_info.sh
```diff
@@ -1,7 +1,24 @@
 # Prints tmux session info.
-# Assuems that [ -n "$TMUX"].
+# Assumes that [ -n "$TMUX"].
+
+TMUX_POWERLINE_SEG_TMUX_SESSION_INFO_FORMAT_DEFAULT='[ #S ]:#I.#P'
```
erikw, line 4: I'd say let's keep the old format as default, to not change unnecessarily for people upgrading by git pull

erikw:
```suggestion
TMUX_POWERLINE_SEG_TMUX_SESSION_INFO_FORMAT_DEFAULT='#S:#I.#P'
```

PR #302, opened 2023-03-25
Add ability to selectively disable spacing and segment characters

henrypotgieter:
This commit allows you to specify additional segment options in your themes file to have more fine control over the rendering of specific segments in the status bar.  The goal was to allow a more seamless integration with the tmux-mem-cpu-load plugin so as to enable nesting it within other segments and still having the powerline look good.  Other custom statusbar plugins may also benefit from this feature.

R = Disk Free segment
G = Slightly modified tmux-mem-cpu-load segment
B = Battery segment

Before:
before

After:
after

Note that the tmux-mem-cpu-load plugin in the screenshots has been modified to allow specification of the background colour in the memory section and to insert a trailing separator character in the load section so that dynamic colours seamlessly blend with the neighbouring segments.

This is an example of the adjustment required in the theme file:

theme_file_changes

The new options are not required to be set, so existing segment function is not impacted.  In the battery segment it was desirable for instance to not strip any leading or tailing whitespace but to only remove the separator at the start of the segment.  Thus an N was used (any character except L,R, or B) to ensure the conditions were not met to remove spacing.

### DIFF
--- a/lib/powerline.sh
+++ b/lib/powerline.sh
@@ -74,6 +74,8 @@ __process_segment_defaults() {
 			${input_segment[1]:-$TMUX_POWERLINE_DEFAULT_BACKGROUND_COLOR} \
 			${input_segment[2]:-$TMUX_POWERLINE_DEFAULT_FOREGROUND_COLOR} \
 			${input_segment[3]:-$default_separator} \
+			${input_segment[6]:-$spacing_disable} \
+			${input_segment[7]:-$separator_disable} \
 		)
 
 		powerline_segments[$segment_index]="${powerline_segment_with_defaults[@]}"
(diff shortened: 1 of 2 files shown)


lib/powerline.sh
```diff
@@ -105,7 +107,15 @@ __process_scripts() {
 		fi
 
 		if [ -n "$output" ]; then
-			powerline_segment_contents[$segment_index]=" $output "
+			if [ ${powerline_segment[4]} == "L" ] ; then
```
erikw, line 110: `L` means 'output to the left of the space,
`R` means 'output to the right of the space,
`B` ?

I'm trying to understand the choice of letters :). It might not be so intuitive looking at a config and knowing what these does. 

And https://github.com/erikw/tmux-powerline/blob/0995e9a37214fae6e94d5c361b2ba73f84efe438/themes/default.sh#L51 would need an update

erikw (changes requested):
Hey @henrypotgieter and thanks for the PR!

I understand the need for more control of the printing for certain segments. I like that you have made it opt-in 👍 .

I have just one comment below. We should explain in the default theme what these options means.

lib/powerline.sh
henrypotgieter: OK all done!  I've improved the naming of the options so they're more easy to understand there.

The themes/default.sh file has been updated to include a pretty thorough usage example and information about the existing and new optional fields.

erikw (approved):
Great, thanks or making the config names more readable and adding the extensive theme format comments!

PR #304, opened 2023-04-16
add variables for users to set mute bindings

TheSast:
Now the user has default keybindings for muting and can change/unbind them with a variable.

From
```
bind M-Right run '$HOME/.config/tmux/plugins/tmux-powerline/mute_powerline.sh right'
```
To
```
set -g @powerline-mute-right 'M-Right'
```
It also should work with different `$TMUX_PLUGIN_MANAGER_PATH`s


erikw:
Hey thanks for the suggestion. I thought about making the keybinding as opt-in configuration,but I cut it from the scope to be able to wrap up in time yesterday.

When I adapted for the side length I faced the question: do I make them configurable the tpm way with tmux user options, or use the existing tmux-powerline configuration file. I think it creates confusion when some settings are in tmux.conf and some in tmux-powerline config. So I decided it's best to have all in one place and thus implemented `TMUX_POWERLINE_STATUS_LEFT_LENGTH` in the config file.

I think we should go this way also for the sidebar mute keybindings :)

TheSast:
I think this type of stuff is best suited for tmux.conf, it is easier to manage keybind configuration when you have them all under your nose.
Furthermore someone that wants to shuffle their bindings (e.g. changing all window related bindings to have Shift in them and their previous muting bind was on a needed key) would have to go and open up the powerline rc file.

To make it an opt-in you can currently just leave the defaults as `no-key`. Otherwise I could remove the fallback if you prefer.

erikw:
I see your point from having an overview of our keybindings from your `tmux.conf`. 

I need to have the perspective of the maintainability of the code and future evolution of it. I would like to avoid supporting two places where configuration happens. Not only for code maintainability reason, but because if there's more than one way to configure, the discussion will always come again and again when someone is contributing a new feature, where this should now be configured. It's easy to avoid having that discussion every time, by having one configuration place. Given that there is already an extensive configuration file for the project, it's more pragmatical to continue using this, even though I see your argument for this particular conf (keybinding).

I implemented it as 2 new configuration variables in f8f22cd96bd3c0b10b3ac641113997f40873f86f:

Adding to `~/.config/tmux-powerline/config.sh`:
```conf
	export TMUX_POWERLINE_MUTE_LEFT_KEYBINDING="C-["
	export TMUX_POWERLINE_MUTE_RIGHT_KEYBINDING="C-]"
```
and they will be picked up and set as keybindings to mute/unmute the status bars :) 

Now the README is also smaller, making it less daunting to look at as a new user. Thanks for the idea and motivation! :)

PR #306, opened 2023-04-26
Config option to set status bar initial visibility, refresh interval and justificaiton

### LINKED ISSUE #305: Changing default tmux-powerline status settings
While I think the default status settings are fine, they are definetly not great.
I'm specifically talking about these:
https://github.com/erikw/tmux-powerline/blob/f8f22cd96bd3c0b10b3ac641113997f40873f86f/main.tmux#L15-L17

1. I know a few people that have their status only show up in certain cases (e.g. prefix pressed, window/session menu) and this would conflict with that setup.
1. The [time](https://github.com/erikw/tmux-powerline/blob/main/segments/time.sh) and [utc_time](https://github.com/erikw/tmux-powerline/blob/main/segments/utc_time.sh) segments can be formatted to show seconds, which would unintuitively refresh only every other second.
1. I think that tmux-powerline can work and look great even when the window list is left or right justified, I personally use it everyday with the former. 
Also before I updated tmux to 3.3a I personally experienced some strange behaviour from `status-justify "centre"` (incorrect and inconsistent alignment on using no segments/semi-static segments) that didn't arise when using `status-justify "absolute-centre"` which I would reccomend as the replacement default.

I think these issues could be resolved by having the option to set these parameter in the rc file.

TheSast:
Resolves #305


erikw (approved):
LGTM 🚀, thank you for your contributions!

PR #310, opened 2023-05-10
Fix: wrong variable tested during init

LuminousXLB:


erikw:
Good spot -- thank you for the fix!

PR #312, opened 2023-05-13
Fix: Empty XDG_CONFIG_HOME not handled correctly

### LINKED ISSUE #311: Empty XDG_CONFIG_HOME is handled incorrectly
According to the [specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables), an empty or unset `XDG_CONFIG_HOME` should be interpreted as `$HOME/.config`. Currently in this case tmux-powerline searches for `/tmux-powerline` instead of `$HOME/.config/tmux-powerline`

bseibold:
Fixes #311

### DIFF
--- a/config/paths.sh
+++ b/config/paths.sh
@@ -1,5 +1,8 @@
 # Paths
 
+if [ -z "$XDG_CONFIG_HOME" ]; then
+	XDG_CONFIG_HOME=$HOME/.config
+fi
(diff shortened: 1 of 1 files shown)


erikw:
Great find! A default value should definitely be used.

I modified the PR slightly; I usually use the default value feature of bash like this for XDG vars: `${XDG_CONFIG_HOME:-$HOME/.config}`

PR #313, opened 2023-05-17
now_playing.sh: bugfix - fixed music grubbing using WSL

itaysaadia:
removed /APPS, it's not needed anymore


erikw:
How come `/APPS` is no longer needed? I see though that it's no longer listed in the latest manual:

https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/tasklist

As I don't have a Windows system at hand I'll trust you on that it works :)

PR #315, opened 2023-06-12
fix(formatting.sh): return any non-abbreviated color as-is

### LINKED ISSUE #314: named colours not rendering
I'd like to be able to specify named colours so that when my terminal theme changes, tmux-powerline will automatically inherit those colours

### Problem

When providing colours for themes, named colours, such as `black | blue | cyan | green | etc.` are not being rendered. 

### Steps to reproduce

- open ~/.config/tmux-powerline/themes/{current-theme}.sh
- go to any segment
- replace background or foreground colour with `red`

### Expected result

The terminal colour set for `red` is applied

### Actual result

Styles fall back to tmux's default for the foreground / background

### Possible solutions

[formatting.sh](https://github.com/erikw/tmux-powerline/blob/main/lib/formatting.sh) accepts 3 types of values:

- a literal `default` 
- hex values in the form `#[xxxxxx]`
- coerced `colo[u]rxxx` values in the form `\d{1,3}` where `colo[u]r` has been omitted from the provided value, and inserted by `tmux-powerline`

This means that `tmux-powerline` incorrectly formats valid colour values of the forms:

- `colo[ur]xxx`, .e.g `colour100` or becomes `colourcolour100`
- `[named color]`, e.g. `red` becomes `colourred`

We can maintain the current coercion of `\d{1,3}` patterned values while allowing all other values to be accepted as-is:

```shell
# lib/formatting.sh
__normalize_color() {
	local input="$1"
	local result

	case "$input" in
			[0-9]|[0-9][0-9]|[0-9][0-9][0-9]) # handle 1 to 3 digits
					result="colour$input"
					;;
			*) # Catch-all
					result=$input
					;;
	esac

	echo -n "$result"
}

__print_colored_content() {
	local bgcolor
	local fgcolor

	bgcolor=$(__normalize_color "$2")
	fgcolor=$(__normalize_color "$3")

	echo -n "#[fg=${fgcolor},bg=${bgcolor}]"
	echo -n "$1"
	echo -n "#[default]"
}
```

larrybotha:
Allows for using named valid colours when theming `tmux-powerline`, e.g. `red`, `color123`, `colour123`, `default`, in addition to converting abbreviated colour values to their `colo[u]r` variants

closes #314

### DIFF
--- a/lib/colors.sh
+++ b/lib/colors.sh
@@ -1,0 +1,31 @@
+# Utilities for working with colors
+#
+# Dependencies:
+#		- none
+
+#######################################
+# Accepts a single argument containing a color, returning a normalized color based
+# on the input
+#
+# If an integer between 1 and 3 characters long is provided, returns `colour[$1]`
+# otherwise returns the input unaltered
+#
+# Arguments:
+#   $1: The color to normalize
+#######################################
+__normalize_color() {
+	local input="$1"
+	local result
+
+	case "$input" in
+		[0-9]|[0-9][0-9]|[0-9][0-9][0-9]) # handle 1 to 3 digits
+			result="colour$input"
+			;;
+		*) # Catch-all
+			result=$input
+			;;
+	esac
+
+	echo -n "$result"
+}
+
--- a/lib/headers.sh
+++ b/lib/headers.sh
@@ -1,6 +1,6 @@
 # Source all needed libs and helpers, kind of like a main.h.
 
-if [ "$TMUX_POWERLINE_DIR_HOME" = "" ]; then
+if [ -z "$TMUX_POWERLINE_DIR_HOME" ]; then
 	lib_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 	TMUX_POWERLINE_DIR_HOME="$(dirname "$lib_dir")" # step up to parent dir.
 
--- a/lib/headers.sh
+++ b/lib/headers.sh
@@ -11,6 +11,7 @@ source "${TMUX_POWERLINE_DIR_HOME}/config/paths.sh"
 source "${TMUX_POWERLINE_DIR_HOME}/config/shell.sh"
 source "${TMUX_POWERLINE_DIR_HOME}/config/defaults.sh"
 
+source "${TMUX_POWERLINE_DIR_LIB}/colors.sh"
 source "${TMUX_POWERLINE_DIR_LIB}/arg_processing.sh"
 source "${TMUX_POWERLINE_DIR_LIB}/formatting.sh"
 source "${TMUX_POWERLINE_DIR_LIB}/muting.sh"
--- a/lib/headers.sh
+++ b/lib/headers.sh
@@ -1,8 +1,10 @@
-# Source all neede libs and helpers, kind of like a main.h.
+# Source all needed libs and helpers, kind of like a main.h.
 
-if [ -z "$TMUX_POWERLINE_DIR_HOME" ]; then
+if [ "$TMUX_POWERLINE_DIR_HOME" = "" ]; then
 	lib_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-	export TMUX_POWERLINE_DIR_HOME="$(dirname "${lib_dir}")" # step up to parent dir.
+	TMUX_POWERLINE_DIR_HOME="$(dirname "$lib_dir")" # step up to parent dir.
+
+	export TMUX_POWERLINE_DIR_HOME
 	unset lib_dir
 fi
 
--- a/lib/powerline.sh
+++ b/lib/powerline.sh
@@ -2,7 +2,9 @@
 
 print_powerline() {
 	local side="$1"
-	local upper_side=$(echo "$1" | tr '[:lower:]' '[:upper:]')
+	local upper_side
+
+	upper_side=$(echo "$1" | tr '[:lower:]' '[:upper:]')
 	eval "local input_segments=(\"\${TMUX_POWERLINE_${upper_side}_STATUS_SEGMENTS[@]}\")"
 	local powerline_segments=()
 	local powerline_segment_contents=()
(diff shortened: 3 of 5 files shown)


lib/headers.sh
```diff
@@ -1,8 +1,10 @@
-# Source all neede libs and helpers, kind of like a main.h.
+# Source all needed libs and helpers, kind of like a main.h.
 
-if [ -z "$TMUX_POWERLINE_DIR_HOME" ]; then
```
erikw, line 3: Bash improvements are appreciated! Though this is not related to fixing #314, thus it would have been best to split this to a separate PR. It would make it easier to review, and also debug if errors are introduced.

Nevertheless, let's continue here. `-z` is not a good idea?

lib/colors.sh
```diff
@@ -0,0 +1,31 @@
...
+# Accepts a single argument containing a color, returning a normalized color based
+# on the input
+#
+# If an integer between 1 and 3 characters long is provided, returns `colour[$1]`
+# otherwise returns the input unaltered
+#
+# Arguments:
+#   $1: The color to normalize
+#######################################
+__normalize_color() {
```
erikw, line 16: 👍 great idea to extract this, makes usage clearer in formatting, and with more features!

lib/headers.sh
```diff
@@ -1,8 +1,10 @@
...
+if [ "$TMUX_POWERLINE_DIR_HOME" = "" ]; then
 	lib_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-	export TMUX_POWERLINE_DIR_HOME="$(dirname "${lib_dir}")" # step up to parent dir.
```
erikw, line 5: I recall shellcheck complains on this

lib/powerline.sh
```diff
@@ -2,7 +2,9 @@
 
 print_powerline() {
 	local side="$1"
```
erikw, line 4: Is shellcheck not complaining about `side` as well?

lib/powerline.sh
```diff
@@ -17,23 +19,28 @@ print_powerline() {
...
-		echo "fg=colour$TMUX_POWERLINE_DEFAULT_FOREGROUND_COLOR,bg=colour$TMUX_POWERLINE_DEFAULT_BACKGROUND_COLOR,nobold,noitalics,nounderscore"
-		;;
-	*)
-		;;
-    esac
+	local type="$1"
+	local bg_color
+	local fg_color
+
+	bg_color=$(__normalize_color "$TMUX_POWERLINE_DEFAULT_BACKGROUND_COLOR")
```
erikw, line 26: Nice extraction

lib/powerline.sh
```diff
@@ -80,13 +93,13 @@ __process_segment_defaults() {
 			${input_segment[7]:-$separator_disable} \
 		)
 
-		powerline_segments[$segment_index]="${powerline_segment_with_defaults[@]}"
+		powerline_segments[segment_index]="${powerline_segment_with_defaults[@]}"
```
erikw, line 96: Why remove `$` here, and below? A spellcheck recommendation? If so, could you please link 🤔

erikw (commented):
Great feature, and nice job with the implementation.

Great with some bash fix-ups, even though I would have appreciate to keep this in a separate PR.

Could drop a line in `CHANGELOG.md` as well about this update?

lib/headers.sh
larrybotha: ye... I have both `shellharden` and `shellfmt` doing some autoformatting... it felt dirty committing these here, but I did the lazy thing instead of the correct thing :/

Apologies, I should have split these into a separate PR

Turns out that this is actually `shellharden` doing this - using `-z` is technically the correct way to do things, but I've read that comparing against an empty string is useful in terms of readability, particularly when someone's unfamiliar with bash

It seems as though `shellharden` is favouring readability over being pedantic

I'll revert if you prefer - I'm neither here nor there about which style to use :)

lib/headers.sh
larrybotha: ye, when values are derived from subprocesses it seems to prefer separating declaration and assignment

lib/powerline.sh
larrybotha: not here, but it was on ln:150 and 152 of the current changes :)

lib/powerline.sh
larrybotha: ye, it seems as though when using a value as an array accessor you don't need the variable O.o

https://www.shellcheck.net/wiki/SC2004

It looks as though it's derived from arithmetic rules, however, which is curious

lib/headers.sh
erikw: I see, I'm not familiar with `shellharden`. I'd say let's revert to `-z` for consistency, as it's used everwhere else in this project :)

lib/powerline.sh
erikw: Oh interesting, thanks for sharing!

lib/headers.sh
larrybotha: done https://github.com/erikw/tmux-powerline/pull/315/commits/e1f4a7a32c2247227c5e3a9c5b56c6d1105db972 :)

larrybotha:
> Great feature, and nice job with the implementation.
> 
> As well with the bash fix-ups, even though I would have appreciate to keep this in a separate PR.
> 
> Could drop a line in `CHANGELOG.md` as well about this update?

It looks like the minor versions are being bumped for bug fixes - should I continue with this approach for adding to the CHANGELOG.md, or should I use the conventional _patch_ bump for the bugfix?

erikw:
@larrybotha just append after line 10 here
https://github.com/erikw/tmux-powerline/blob/18b4fd6e4f43911e04a1912a1daa49a661ee79d7/CHANGELOG.md?plain=1#L10

larrybotha:
@erikw done :)

I added it under _Fixes_, as I expected named values to work by default, and https://github.com/erikw/tmux-powerline/blob/main/themes/default.sh#L53-L64 also describes using named colours

Let me know if you'd prefer it under _Added_

erikw (approved):
Many thanks, great PR!

PR #316, opened 2023-06-19
Add VCS branch truncation

embe221ed:
I've added the truncation mechanism to the `vcs_branch` segment in order to make longer branches' names look nicer.

### DIFF
--- a/segments/vcs_branch.sh
+++ b/segments/vcs_branch.sh
@@ -62,7 +62,7 @@ __parse_git_branch() {
 
 	# Clean off unnecessary information.
 	branch=${branch#refs\/heads\/}
-  branch=$(__truncate_branch_name $branch)
+	branch=$(__truncate_branch_name $branch)
 
 	echo -n "#[fg=colour${git_colour}]${branch_symbol} #[fg=colour${TMUX_POWERLINE_CUR_SEGMENT_FG}]${branch}"
 }
@@ -84,7 +84,7 @@ __parse_svn_branch() {
 	local svn_url=$(echo "${svn_info}" | sed -ne 's#^URL: ##p')
 
 	local branch=$(echo "${svn_url}" | grep -E -o '[^/]+$')
-  branch=$(__truncate_branch_name $branch)
+	branch=$(__truncate_branch_name $branch)
 	echo "#[fg=colour${svn_colour}]${branch_symbol} #[fg=colour${TMUX_POWERLINE_CUR_SEGMENT_FG}]${branch}"
 }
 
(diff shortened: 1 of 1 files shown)


segments/vcs_branch.sh
```diff
@@ -50,6 +52,8 @@ __parse_git_branch() {
 
 	# Clean off unnecessary information.
 	branch=${branch#refs\/heads\/}
+	trunc_symbol="···"
+  branch=$(echo $branch | sed "s/\(.\{$TMUX_POWERLINE_SEG_VCS_BRANCH_MAX_LEN_DEFAULT\}\).*/\1$trunc_symbol/")
```
erikw, line 56: Indentation looks off

segments/vcs_branch.sh
```diff
@@ -3,6 +3,8 @@
 # Source lib to get the function get_tmux_pwd
 source "${TMUX_POWERLINE_DIR_LIB}/tmux_adapter.sh"
 
+TMUX_POWERLINE_SEG_VCS_BRANCH_MAX_LEN_DEFAULT="24"
```
erikw, line 6: Good idea to make it configurable.

A few things are missing from this follow the pattern for configuration in this project:
* Have a `_DEFAULT`, like already done here
* in the `generate_segmentrc()` function (missing in this file, needs to be added), add code to generate a default config file with default value.
* In `__process_settings()` (missing here, needs to be added) set the value or use default value.
* In the `rung_segment()`, use not the `_DEFAULT` but the set value.

It sounds way more complicated than it is. Take a look at the file https://github.com/erikw/tmux-powerline/blob/main/segments/battery.sh and ctrl/cmd+f for `TMUX_POWERLINE_SEG_BATTERY_NUM_HEARTS` to see how it's done, then just copy this same way here :)

erikw:
Great idea, just two comments

segments/vcs_branch.sh
```diff
@@ -86,5 +100,20 @@ __parse_hg_branch() {
 	fi
 
 	local branch=$(echo "$summary" | grep 'branch:' | cut -d ' ' -f2)
-	echo  "#[fg=colour${hg_colour}]${branch_symbol} #[fg=colour${TMUX_POWERLINE_CUR_SEGMENT_FG}]${branch}"
+	branch=$(__truncate_branch_name $branch)
+	echo "#[fg=colour${hg_colour}]${branch_symbol} #[fg=colour${TMUX_POWERLINE_CUR_SEGMENT_FG}]${branch}"
+}
+
+
+__truncate_branch_name() {
```
erikw, line 108: Even better, to make use of the truncation for SVN as well!

erikw (approved):
Great, thank you for this addition!

PR #318, opened 2023-06-23
Remove extra S from custom segments section

akan72:
## Summary
- Removing extra "s" character from `Custom Segments` in README

## Test Plan
Markdown renders without typo
image


erikw:
Thx!

PR #321, opened 2023-07-06
Palette has 8 colors in a line

tnkhanh:
I can't have all colors fit on my screen if there are 4 colors on a line.
It's better to have 8 colors on a line so we don't have to scroll to compare all colors.


erikw:
Good idea, thanks!

PR #324, opened 2023-07-22
[weather] mapping more weather codes on https://api.met.no/weatherapi/weathericon/2.0/documentation 

deanbear:
This PR includes the following optimizations:

1. I have mapped additional weather codes from https://api.met.no/weatherapi/weathericon/2.0/documentation and differentiated some of them for day/night variations.
2. I have replaced part of icons with emojis because the 'clear sky' symbol was not displaying correctly on my Mac OS system.
3. To resolve the issue with the configuration not taking effect, I utilized the TMUX_POWERLINE_SEG_WEATHER_UPDATE_PERIOD from config.sh to set the interval for calling the weather API.

Have a wonderful day.

### DIFF
--- a/segments/weather.sh
+++ b/segments/weather.sh
@@ -1,7 +1,4 @@
-# Prints the current weather in Celsius, Fahrenheits or lord Kelvins. The forecast is cached and updated with a period of $update_period.
-
-# The update period in seconds.
-# update_period=600
+# Prints the current weather in Celsius, Fahrenheits or lord Kelvins. The forecast is cached and updated with a period.
 
 TMUX_POWERLINE_SEG_WEATHER_DATA_PROVIDER_DEFAULT="yrno"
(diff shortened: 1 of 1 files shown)


segments/weather.sh
```diff
@@ -1,7 +1,7 @@
 # Prints the current weather in Celsius, Fahrenheits or lord Kelvins. The forecast is cached and updated with a period of $update_period.
 
 # The update period in seconds.
-update_period=600
+# update_period=600
```
erikw, line 4: Hm it looks like this one is not used anymore, right? If so we could remove these 2 lines.

erikw (changes requested):
Thank you, great updates!

Just one clean-up we could do :)

segments/weather.sh
deanbear: roger that~

erikw (approved):
Thank you Bear!

Issue #331, opened 2023-09-21
Battery segment not displaying


skywarth:
Hi,

I just installed tmux-powerline and was about to customize it. So far all the segments defined in the default configs seems to be working, except for battery segment. It simply doesn't display it at all. I disabled all other segments to test maybe it wasn't able to fit in the stack width, that didn't change anything. Screenshot below, certain fields are hidden:

BTW: default tmux (without tmux-powerline plugin) displays the battery just fine. 

![Screenshot from 2023-09-22 01-38-58](https://github.com/erikw/tmux-powerline/assets/23261263/67410c9a-7b6e-45d6-87b1-1e1b96450a54)

## My configuration (basically default config):

```
# Default configuration file for tmux-powerline.
# Modeline {
#	 vi: foldmarker={,} foldmethod=marker foldlevel=0 tabstop=4 filetype=sh
# }

# General {
	# Show which segment fails and its exit code.
	export TMUX_POWERLINE_DEBUG_MODE_ENABLED="false"
	# Use patched font symbols.
	export TMUX_POWERLINE_PATCHED_FONT_IN_USE="true"

	# The theme to use.
	export TMUX_POWERLINE_THEME="my-theme"
	# Overlay directory to look for themes. There you can put your own themes outside the repo. Fallback will still be the "themes" directory in the repo.
	export TMUX_POWERLINE_DIR_USER_THEMES="${XDG_CONFIG_HOME:-$HOME/.config}/tmux-powerline/themes"
	# Overlay directory to look for segments. There you can put your own segments outside the repo. Fallback will still be the "segments" directory in the repo.
	export TMUX_POWERLINE_DIR_USER_SEGMENTS="${XDG_CONFIG_HOME:-$HOME/.config}/tmux-powerline/segments"

	# The initial visibility of the status bar. Can be {"on, off"}.
	export TMUX_POWERLINE_STATUS_VISIBILITY="on"
	# The status bar refresh interval in seconds.
	# Note that events that force-refresh the status bar (such as window renaming) will ignore this.
	export TMUX_POWERLINE_STATUS_INTERVAL="1"
	# The location of the window list. Can be {"absolute-centre, centre, left, right"}.
	# Note that "absolute-centre" is only supported on `tmux -V` >= 3.2.
	export TMUX_POWERLINE_STATUS_JUSTIFICATION="centre"

	# The maximum length of the left status bar.
	export TMUX_POWERLINE_STATUS_LEFT_LENGTH="60"
[...]

PR #334, opened 2023-09-28
Alternated Battery Status Condition Check in Bash

skywarth:
### Overview
This PR addresses an issue with the battery status condition check in the Bash function. Under certain conditions, particularly when the input parameter was an empty string, the previous condition regarding comparison to `cat $STATUS` using `[ ... ]` could become malformed.

**I can develop some additional conditions and checks for compatibility with POSIX-compliant shells if it's a target platform.**

### Linked issue/bug report

Please see #331 for additional details on the bug and findings.

### Changes
- Switched to using the `[[ ... ]]` test construct for the battery status check. This provides a more robust and predictable behavior in Bash.

### Impact
- Users with Bash as their shell will experience a more reliable battery status check.
- This change is Bash-specific and may not be portable to strictly POSIX-compliant shells.
- Be wary that it could break compatibility with other shell variants such as `sh`, `ash`, `dash` perhaps. Haven't tested it in those shells.

### Testing
The changes have been tested on a default Ubuntu 22.04 installation with no shell customization. Only tmux, tpm and tmux-powerline is installed for terminal. 

Please review and let me know if there are any concerns or feedback.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -37,3 +37,4 @@ MSempere
 kurczynski
 Alejandro Espinosa (Halbard) <alejosp493@gmail.com>
 TN Khanh <tnkhanh4@gmail.com>
+Skywarth <yigitk.ersoy@gmail.com>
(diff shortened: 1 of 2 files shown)


erikw:
Thanks, bash is a requirement. 
https://github.com/erikw/tmux-powerline#requirements

PR #335, opened 2023-10-01
Remove support for ~/.tmux-powerlinerc

### LINKED ISSUE #330: README.md out of date?
Section Config: https://github.com/erikw/tmux-powerline#config-file

seems to be wrong, or at least out of date. I have zero prior knowledge of tmux-powerline (nor tpm), but i barely made it working. Now, the problem seems to be I cannot configure it. Despite trying to follow the config section in the README.md, which is partly wrong, I feel like my changes to `~/.tmux-powerlinerc` are ignored.

Most importantly, the segments are showing way to much, like everything (so it feels), so I tried to find not to display that much, like, disabling some things, like ifstat.sh, but I seem not to find the documentation that tells me how to. - Do you have any hints for me?

Many thanks.

erikw:
This path has been deprecated for a while. It seems to cause some issues
in #330, so it's maybe time to remove it.

Create a new major release after this is merged.

Fixes #330


benjaminmikiten:
Thanks for the quick response to #330 !

erikw:
Tested here: https://github.com/erikw/tmux-powerline/issues/330#issuecomment-1742315137

PR #337, opened 2023-10-15
Add Plexamp support

lunks:
This adds Plexamp support similar to how we handle Last.fm.
Plex API uses XML which would bring another dependency, so I'm using Tautulli
to grab the now playing info.


erikw:
Thanks for detailed instructions!

PR #338, opened 2023-10-21
add segment for gcalcli

Drew-Daniels:
I've been using the [gcalcli](https://github.com/insanum/gcalcli) tool since yesterday, which allows you to perform basic Google Calendar actions from your command line, and they had a [recommendation](https://github.com/insanum/gcalcli#agenda-integration-with-tmux) on how you could use the output from running `gcalcli agenda | head -2 | tail -1` to print your next Google Calendar event on your `tmux` status line:

image

I thought this was pretty cool, and hacked something together to display the output from this command as a `tmux-powerline` segment. The main updates I've made are:
- Adding a `TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME` option so users can opt out of using military time. 
  - When 0, the `--no-military-time` option will be provided to `gcalcli`
  - When 1 (Default), the `--no-military-time` option _will not_ be provided to `gcalcli`
- Stripping all color coding information from the standard output of `gcalcli agenda`, which usually looks like this:

image

- Stripping extra whitespace from the output so that it does not unnecessarily take up more space than necessary to display events.

This would be my first open source contribution so if there's anything I've missed, please let me know! But thought other people might get some use out of this.

Here's what it looks like without any fancy text/background colors:

`TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME == 1` (Default)

image

`TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME == 0`

image

### DIFF
--- a/segments/gcalcli.sh
+++ b/segments/gcalcli.sh
@@ -1,0 +1,32 @@
+# Print next Google Calendar Event
+
+TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME_DEFAULT=1
+
+generate_segmentrc() {
+	read -d '' rccontents  << EORC
+gcalcli uses military time by default - if you want to see 12 hour time format, set TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME_DEFAULT to 0
+export TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME="${TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME_DEFAULT}"
+EORC
+	echo "$rccontents"
+}
+
+__process_settings() {
+	if [ -z "$TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME" ]; then
+		export TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME="${TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME_DEFAULT}"
+	fi
+}
+
+run_segment() {
+  if ! command -v gcalcli &> /dev/null
+  then
+    echo "'gcalcli' could not be found"
+    return 1
+  fi
+  __process_settings
+  gcmd=(gcalcli agenda)
+  if [[ $TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME == 0 ]]; then
+    gcmd+=(--no-military)
+  fi
+  "${gcmd[@]}" | head -2 | tail -1 | sed 's/\x1B\[[0-9;]\{1,\}[A-Za-z]//g' | sed -E 's/ +/ /g'
+	return 0
+}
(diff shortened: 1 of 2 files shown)


segments/gcalcli.sh
```diff
@@ -0,0 +1,32 @@
+# Print next Google Calendar Event
+
+TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME_DEFAULT=1
+
+generate_segmentrc() {
+	read -d '' rccontents  << EORC
+gcalcli uses military time by default - if you want to see 12 hour time format, set TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME_DEFAULT to 0
```
erikw, line 7: Prepend ‘#’ to make this a comment, the generated file is a shell script:)

segments/gcalcli.sh
```diff
@@ -0,0 +1,32 @@
...
+export TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME="${TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME_DEFAULT}"
+EORC
+	echo "$rccontents"
+}
+
+__process_settings() {
+	if [ -z "$TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME" ]; then
```
erikw, line 14: I would change military time to “24hr” or similar, as I’m quite sure this is more common term to use (in the software business at least).

segments/gcalcli.sh
```diff
@@ -0,0 +1,32 @@
...
+		export TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME="${TMUX_POWERLINE_SEG_GCALCLI_MILITARY_TIME_DEFAULT}"
+	fi
+}
+
+run_segment() {
+  if ! command -v gcalcli &> /dev/null
```
erikw, line 20: Thanks for checking runtime dependencies!

erikw (changes requested):
Great idea, I’ve used this tool in the past. 

Thank you for your (first) contribution! Some small updates needed, then we can get this merged:)

segments/gcalcli.sh
Drew-Daniels: Ah thanks for catching that! Totally missed that.

segments/gcalcli.sh
Drew-Daniels: That's a good point, I've updated this 👍🏻

Drew-Daniels:
> Great idea, I’ve used this tool in the past.
> 
> Thank you for your (first) contribution! Some small updates needed, then we can get this merged:)

Sure thing! Thanks for the great feedback. Just pushed up some updates - let me know if there's anything else I need to update

erikw (approved):

erikw:
Thank you!

PR #339, opened 2023-10-30
Do not create RCDIR if it's a symlink

joona:
My RCDIR is linked to another directory, inside my `dotfiles` repository. Currently this results to TPM failing with error 1 when reloading config.

Adds detection for RCDIR symlink and only creates the directory, if it's not a directory or symlink.


erikw:
Makes sense, thanks!

Issue #344, opened 2023-11-06
sharing my config and battery segment


embe221ed:
Hey,

I wanted to share few things that I've changed in my config as I find them pretty good.
I'll prepare two PRs for these in case you find it good.

1. changed the `battery.sh` script to improve the icons displayed
the heart icon is replaced with battery/plug icon to better reflect the current state. I've changed the color of the icon to red in case the battery level is lower than 50%.
The most improved version is for MacOS, but I've also done some improvements to the Linux version.

2. my "bubble" theme
replaced separators, updated colors, for example Spotify colors are the same as in Spotify logo, Github - as well.

few screenshots:
1. charging
<img width="2048" alt="Screenshot 2023-11-06 at 12 13 00" src="https://github.com/erikw/tmux-powerline/assets/26097838/a72d0429-1fab-421c-8523-21a38f9846dc">

2. not charging, 100-80%
<img width="2048" alt="Screenshot 2023-11-06 at 12 13 11" src="https://github.com/erikw/tmux-powerline/assets/26097838/1e8cdab0-1934-442f-89ae-f9e3605bdcea">

3. not charging, 80-50%
<img width="2048" alt="Screenshot 2023-11-06 at 12 14 20" src="https://github.com/erikw/tmux-powerline/assets/26097838/54f9bc9b-4fe2-4f7e-b46f-c7e9b586401c">

4. not charging, less than 50%
<img width="2048" alt="Screenshot 2023-11-06 at 12 13 58" src="https://github.com/erikw/tmux-powerline/assets/26097838/7bb9a7af-7e75-487c-a6b6-1b383db993cb">

erikw:
Hey, those sounds like cool improvements :)

marsjane:
I tried yours config, but when I reload config, it shows "'~/.tmux/plugins/tpm/tpm' returned 1", any idea how to debug, where is the error info?

marsjane:
Really cool!

marsjane:
btw, how to set the background of tmux same with the status bar background?

embe221ed:
it depends on the terminal emulator you use. I am using kitty, and here is the repo for that theme: https://github.com/catppuccin/kitty
(kitty/macchiato is used on the screenshots)

marsjane:
[...]

PR #346, opened 2023-11-07
Battery script improved

embe221ed:
as mentioned in #344. This is my improved version of the battery segment


erikw:
Looks great! thank you =)

PR #348, opened 2023-11-15
Support specifying command-line arguments for tmux-mem-cpu-load segment and config gen

rjahanbakhshi:


segments/tmux_mem_cpu_load.sh
```diff
@@ -1,9 +1,17 @@
 # Print out Memory, cpu and load using https://github.com/thewtex/tmux-mem-cpu-load
 
+generate_segmentrc() {
+	read -d '' rccontents  << EORC
+# Arguments passed to tmux-mem-cpu-load                                                                                                                                                                  
+export TMUX_POWERLINE_SEG_TMUX_MEM_CPU_LOAD_ARGS="-v"                                                                                                                                                    # The data provider to use. Currently only "yahoo" is supported.
```
erikw, line 6: The comment looks copied from elsewhere

segments/tmux_mem_cpu_load.sh
```diff
@@ -1,9 +1,17 @@
...
+EORC
+	echo "$rccontents"
+}
+
 run_segment() {
 	stats=""
 	if type $TMUX_PLUGIN_MANAGER_PATH/tmux-mem-cpu-load/tmux-mem-cpu-load > /dev/null 2>&1; then
-		stats=$($TMUX_PLUGIN_MANAGER_PATH/tmux-mem-cpu-load/tmux-mem-cpu-load)
+		stats=$($TMUX_PLUGIN_MANAGER_PATH/tmux-mem-cpu-load/tmux-mem-cpu-load $TMUX_POWERLINE_SEG_TMUX_MEM_CPU_LOAD_ARGS)
```
erikw, line 14: While this works if unset, it would be good practice to let the variable be set but empty by default.

erikw (changes requested):
Good idea. Two comments:)

segments/tmux_mem_cpu_load.sh
rjahanbakhshi: My bad, fixed.

segments/tmux_mem_cpu_load.sh
rjahanbakhshi: > While this works if unset, it would be good practice to let the variable be set but empty by default.

My bad. Fixed and added more information.

erikw (approved):
Thanks!

PR #350, opened 2023-11-23
Alternative timezone support for time segment

abrahamnm:
## Summary
Added support to display a different timezone than the system default in time.sh segment.


erikw:
Nice add on!

PR #353, opened 2024-01-05
fix: Quote path variables and avoid `-o`/`-a` flags to `[`

hyperupcall:
This fixes a bug in which paths (which could contain whitespace) were not properly quoted. This also avoids using [certain test flags](https://www.shellcheck.net/wiki/SC2166) due to their ambiguity in some cases.


erikw (approved):

erikw:
Thank you!

PR #354, opened 2024-01-21
Created a new segment to track Air Quality "Air"

gavink97:
Air allows users to track air quality in a give location via the Open Weather API.

It utilizes US AQI which is the US standard to measure Air quality and returns different colors + symbols for each index level.

Users can use the TMUX_POWERLINE_SEG_AIR_COLOR variable to have the segment change colors as the Air Quality goes between different index levels.


lib/headers.sh
```diff
@@ -19,3 +19,5 @@ source "${TMUX_POWERLINE_DIR_LIB}/formatting.sh"
 source "${TMUX_POWERLINE_DIR_LIB}/muting.sh"
 source "${TMUX_POWERLINE_DIR_LIB}/powerline.sh"
 source "${TMUX_POWERLINE_DIR_LIB}/rcfile.sh"
+
```
erikw, line 22: Segments should not be sourced here. Please remove and verify that the segment still works :).

erikw (changes requested):
Wow cool! Just ones thing to fix

gavink97:
> Wow cool! Just ones thing to fix

Okay I fixed it 👍🏼

erikw (approved):

erikw:
Thx

PR #360, opened 2024-03-11
Remove noisy air segment logs during `generate_rc.sh`

izzygomez:
These changes were added in this [pull request](https://github.com/erikw/tmux-powerline/pull/354/files). Two quick fixes:
- Add check that `$air_temp_file` exists. If it doesn't, `awk` complains; see screenshot below.
- Remove unnecessary `echo`. 

image

### Testing
Verified that `.generate_rc.sh` runs without noisy logs after changes.

image


erikw:
Thanks!

Issue #322, opened 2023-07-06
Reload after modifying my theme


tnkhanh:
How do I reload after modifying my theme? Sourcing `.tmux.conf` again does not help.
So `tmux-powerline` has `set-hook -g sessions-created` to `init`. It only reloads if I create a new session, That's a bit awkward.
Is there a faster way to reload?
Edit: It is `session-created` (singular)!

tnkhanh:
> 
> I can't see that this plugin set `sessions-created` anywhere. Where did you see this?

@erikw  It's in the file you show me, lol [main.tmux](https://github.com/erikw/tmux-powerline/blob/d75110f3568bcca2c6e1ebcdafd978fbe7d1dbda/main.tmux#L33C16-L33C16)

I think something resets on `session-created` because this happens to me:
* Change `TMUX_POWERLINE_DEFAULT_BACKGROUND_COLOR` or `TMUX_POWERLINE_DEFAULT_FOREGROUND_COLOR` in your theme file
* The change does not take effect until you do `<Prefix>:new` to create a new session

erikw:
> So `tmux-powerline` has `set-hook -g sessions-created`

Ah sry I was lazy grep'd on the misspelling here, `sessions-created`, and thus did not find it. I see, `session-created` is used indeed!


> There is no need for a reload. On every status bar re-draw the plugin is run again from scratch. The default is to re-draw the status bar every second, unless this is changed in your `tmux.conf` (set [here](https://github.com/erikw/tmux-powerline/blob/main/main.tmux#L16), [default value](https://github.com/erikw/tmux-powerline/blob/a62d4f132592753c2a2e6cd69b2a3983d9c7c8b5/config/defaults.sh#L8))

This should still be true; every status interval (1 second), tmux will just execute the powerline script again and the latest version of the tmux-powerline configuration file and theme will be used. 

If you disable a segment in the theme and save the file, it should disappear within 1 second, unless this is changed in your tmux.conf (set [here](https://github.com/erikw/tmux-powerline/blob/main/main.tmux#L16), [default value](https://github.com/erikw/tmux-powerline/blob/a62d4f132592753c2a2e6cd69b2a3983d9c7c8b5/config/defaults.sh#L8))

If the segment disappears, we know the theme is properly reloaded :) 

When I change `TMUX_POWERLINE_DEFAULT_BACKGROUND_COLOR` in my theme, it changes almost directly (<1sec) in the status bar.

[...]

PR #361, opened 2024-03-11
Standardize on "generate config" over "generate rc".

izzygomez:
Support for `~/.tmux-powerlinerc` was removed in [this commit](https://github.com/erikw/tmux-powerline/commit/ebd28233b36b552a8d40e0f267d217ae39b5f1d8), so it might be best to standardize on wording around generation of `config.sh` file. Chose not to update naming for segment functions (i.e. `generate_segmentrc()`) for sake of of keeping this PR small.

_Quick note on motivation for this small change_: I am working on a larger change that hopes to address https://github.com/erikw/tmux-powerline/issues/322 (& related issues). While working on that & trying to make sense of the codepaths in this repo, it was a bit confusing to see the `config.sh` file being referenced as both an "RC file" & a "config file", so felt updating the wording was a nice (albeit small) change.

### Testing
Ran `generate_config.sh` & verified that generated `config.sh.default` file looks as expected. 

image


erikw:
Makes sense to me. The diff looks alright, however I’m unable to actually test run the code at the moment. Would you be available to do some bug fixes in case someone reports it breaking something?

izzygomez:
Sounds good, happy to do bug fixes if this breaks anything, just tag me in issue.

Issue #358, opened 2024-02-21
Powerline init not executed on first session creation


StopMotionCuber:
Hey, I would like to utilise the fancy window list, but that doesn't work right now for me, as it seems the 
`tmux set-hook -g session-created "run-shell '${TMUX_POWERLINE_DIR_HOME}/powerline.sh init'"` doesn't seem to run on initial session creation.

## Steps to reproduce
Assumption: `tmux-powerline` is configured properly, according to README
1. Close all tmux sessions
2. Start a new tmux session with `tmux`
3. Wait for powerline to load

## Expected result
The window list is not the default window list and has a powerline style

## Actual result
The window list has the default style

## Workaround
Starting a _second_ session leads to the expected result.

## Further notes
I think this is caused by the `session-created` hook running _before_ loading all the plugins and stuff. It would be fixed by adding `${TMUX_POWERLINE_DIR_HOME}/powerline.sh init` to the end of the `main.tmux` file. I'm not sure whether there would still be a reason for keeping the hook then.
I also don't understand the reason to go for a hook here in the first place.

erikw:
Fixed by https://github.com/erikw/tmux-powerline/pull/362#issuecomment-1992618595

Issue #356, opened 2024-01-26
Changing background color of the tmux window list


Luh-code:
I can't figure out how to change the background color of the tmux window list in the middle. It is stuck on this green color, it kind of looks like a color from gruvbox, which is the theme I tried configuring the bar to, but I don't think I set it to be that anywhere...

Here's a gist of my theme, which is just a modification of the default: https://gist.github.com/Luh-code/e2135c02ce8fa8f6d33aa780654d2744

And here's how it looks like currently: ![2024-01-26_14_13_16](https://github.com/erikw/tmux-powerline/assets/60015267/bde541b9-1d40-46f3-8a03-13d5c0e74b7d)

Luh-code:
Update: Still haven't figured it out. I've re-riced my whole system since this post, but the problem still persists. Also it is a thing on all themes. Including default and bubble.
![image](https://github.com/erikw/tmux-powerline/assets/60015267/87f90cab-24bc-452a-bf14-88d77924e260)

erikw:
Fixed by https://github.com/erikw/tmux-powerline/pull/362#issuecomment-1992618595

Issue #336, opened 2023-10-03
Change window status color on activity?


mwpardue:
I'm sorry, I'm sure I'm missing something simple, but I'm trying to figure out how to change the colors of a window's status label when there is activity. I don't want to just invert it but rather change to entirely different colors. Ultimately I want to change the activity icon as well, but that's another issue.

I can get the colors to change by doing the below:
```
if [ -z $TMUX_POWERLINE_WINDOW_STATUS_FORMAT ]; then
	TMUX_POWERLINE_WINDOW_STATUS_FORMAT=(
    "#{?window_activity_flag,#[fg=colour16 bg=colour88],#[$(format regular)]}" \
		"  #I#{?window_flags,#F, } " \
		"$TMUX_POWERLINE_SEPARATOR_RIGHT_THIN" \
		" #W "
	)
fi
```

What I can't figure out is why this doesn't color the bold separators at either end. Instead of getting the pretty status bar label changed from end to end, I just get the text label changed and the separators are colored the default color of the rest of the status bar See below for an example. 

<img width="405" alt="image" src="https://github.com/erikw/tmux-powerline/assets/88067650/14f71ce4-63a8-4b2d-9c82-027c1c01fa3c">

Can somebody please help me figure out what I'm doing wrong? I've been working on this long enough that I'm sure I'm repeating myself now and just need at least a nudge in the right direction. 

Thanks in advance for any help anybody can provide!!

erikw:
Hey, exactly where do you put these changes?

erikw:
Hmm I wonder if this is also a current limitation similar to

https://github.com/erikw/tmux-powerline#changign-tmux_powerline_default_ground_color-does-not-have-effect

With my own testing, by restarting the tmux session, changes to `TMUX_POWERLINE_WINDOW_STATUS_FORMAT` did take effect.

mwpardue:
> This seems to be a current limitation similar to
> 
> https://github.com/erikw/tmux-powerline#changign-tmux_powerline_default_ground_color-does-not-have-effect
> 
> With my own testing, by restarting the tmux session, changes to `TMUX_POWERLINE_WINDOW_STATUS_FORMAT` did take effect.

[...]

PR #276, opened 2021-12-16
Automatically powerline-ify the window status elements

stephenprater:
Powerline-ifying the window status elements in the tmux status bar requires me to type obnoxious and difficult to read tmux formatting commands.  Introduce a `powerline.sh init` command which sets these options on connect - ensuring my tmux status bar is super cool and I can impress all of my nerd friends.

Screen Shot 2021-12-16 at 12 12 49 AM

PR #362, opened 2024-03-11
Fix tmux-powerline window status elements rendering & updating. 

izzygomez:
#### Overview 

This PR is a first attempt to address the old/closed issues https://github.com/erikw/tmux-powerline/issues/322 & https://github.com/erikw/tmux-powerline/issues/336, as well as the new/open issues https://github.com/erikw/tmux-powerline/issues/356 & https://github.com/erikw/tmux-powerline/issues/358. 

The problem can roughly be summarized as this: `tmux` window status elements were not being set as expected when `tmux` first initialized (i.e. the options `window-status-current-format`, `window-status-format`, & `status-style` weren't being set) or when `.tmux.conf` was reloaded. 

#### Details

The workaround described in `README.md` before these changes was to reload the theme by creating a new session. This is due to how powerline window status elements were implemented in https://github.com/erikw/tmux-powerline/pull/276 & in https://github.com/erikw/tmux-powerline/commit/e6194b0b7f839dc4eb590a61824ee9f52e7b7cab. Specifically, the old codepath of `powerline.sh init` was only run via `tmux`'s `session-created` hook. As the name suggests, this hook only runs when a new session is created, but this doesn't apply when `tmux` is first started (i.e. hook would not run if following ran: `tmux kill-server; tmux new -s temp`). Unfortunately for us, [`tmux` man pages on hooks](https://man7.org/linux/man-pages/man1/tmux.1.html#HOOKS) confirm that there's no other adequate hook to replace `session-created` with in order to get our desired functionality for free out-of-the-box.

My solution is to update `powerline.sh` & `lib/powerline.sh` to support setting `tmux`'s `window-status-*` options dynamically in the same manner `status-{left,right}` had been previously set. This is done via a straightforward splitting up & renaming of functions in `lib/powerline.sh` to more modular functions to be called directly from `main.tmux`. This exact approach didn't work for the `status-style` option (because user theme isn't loaded in `main.tmux` or `lib/powerline.sh`), so I opted to define that dynamically in `lib/config_file.sh` alongside similar flags.

Another small misc addition is to expose the `window-status-separator` option via `TMUX_POWERLINE_WINDOW_STATUS_SEPARATOR{_DEFAULT}`, which can set by users in their `config.sh` for further window status customization. 

#### Potential issues

As mentioned in [this comment](https://github.com/erikw/tmux-powerline/issues/322#issuecomment-1625300533), this new change does indeed override custom themes set from outside tmux-powerline. I'd argue this is a feature, not a bug — tmux-powerline is the appropriate plugin to use for customizing status bar UI/UX options, & if it isn't nicely interoperable with (e.g.) [tmux-colors-solarized](https://github.com/seebi/tmux-colors-solarized) or other similar plugins, that's OK (imo) & only one should be used exclusively. Happy to hear more thoughts on this front.

#### Testing

Verified that tmux window list is rendered correctly through all these steps
- when first starting `tmux` (`tmux kill-server; tmux new -s temp1`)
- when we detach & re-attach (\<prefix\>+D followed by `tmux a -dt temp1`)
- when we create a new session (\<prefix\>+D followed by `tmux new -s temp2`)
- when `.tmux.conf` is reloaded (`tmux source-file ~/.tmux.conf`)

Also verified that if `TMUX_POWERLINE_DEFAULT_{BACKGROUND,FOREGROUND}_COLOR` are updated in active theme file & `.tmux.conf` is reloaded, the window list UI is re-rendered as expected with no tmux-server restart; if `TMUX_POWERLINE_STATUS_INTERVAL` is set appropriately, all tmux options except `status-style` update automatically without a `.tmux.conf` reload.


erikw:
Wow great, thank you for tackling this bug!

Same as the previous PR; I’m unable to test this  in the reasonable near future. Would you be up for potential follow-up fix PRs if someone reports issues? :)

izzygomez:
Yup! Feel free to tag me on any issues that come up.

PR #363, opened 2024-03-13
Update mode indicator segment for modern usage.

izzygomez:
#### Overview

Rewrote the `mode_indicator.sh` segment from scratch to expose settings to users & make more idiomatic to how segments are now written — seems like it was [written in 2017](https://github.com/erikw/tmux-powerline/pull/247/files) & hadn't been updated since then.

#### Screenshots

Example screenshots with default values, i.e. how it would look out of the box with generated `config.sh` & using default theme:

- _normal mode, mouse mode on_
image

- _prefix mode, mouse mode on_
image

- _normal mode, mouse mode on, in copy mode_
image

Example screenshots with custom values with custom theme, e.g `normal`→`•`, `prefix`→`^`, `mouse`→`🐭`, `copy`→`📋`, `separator`→`|`:

- _normal mode, mouse mode on, in copy mode_
image

- _prefix mode, mouse mode disabled_
image

...etc, I think this captures the gist, this segment is now very customizable.

#### Testing
- Verified that `config.sh.default` is generated with new section as expected after `generate_config.sh`.
- Verified that all newly exposed settings in `config.sh` have correct default behavior, & can be overwritten as expected.
- Verified that background & foreground colors set through theme are respected (if not overridden in `config.sh`).
- Verified that minor changes to `lib/powerline.sh` & `segments/vcs_{branch,modified}.sh` didn't break anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Some examples of segments available that you can add to your tmux status bar are
 * Date and time
 * Hostname
 * tmux info
-# tmux mode indicator (normal/prefix, mouse, copy modes)
+* tmux mode indicator (normal/prefix, mouse, copy modes)
 * CWD in pane
 * Current X keyboard layout
 * Network download/upload speed
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Some examples of segments available that you can add to your tmux status bar are
 * Date and time
 * Hostname
 * tmux info
+# tmux mode indicator (normal/prefix, mouse, copy modes)
 * CWD in pane
 * Current X keyboard layout
 * Network download/upload speed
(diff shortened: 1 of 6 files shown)


izzygomez:
_Side note:_ this'll likely be the last PR I'll send over for a while after the few I sent over last couple days — I just happened to be motivated to hack on this because I've been in a tmux/dotfiles hacking mood :). Thanks for reviewing & accepting my changes! As with previous PRs that I made to project, happy to be the point of contact if any issues arise related to changes I'm making here.

erikw:
Looks great! Thanks for including screenshots and testing notes

PR #364, opened 2024-03-13
Set `message-style` to also be default powerline style.

izzygomez:
Like `status-style`, which sets the color scheme of the entire status line (added in https://github.com/erikw/tmux-powerline/pull/362), `message-style` sets the color scheme of the command/prompt mode (i.e. when you press `Prefix` + `:`). This makes the status line color experience even more consistent.

Screenshots:
_Before_
image

_After_
image

[Re this comment](https://github.com/erikw/tmux-powerline/pull/363#issuecomment-1993622395): ok _this_ is the last PR 😅


erikw:
💡 thx!

PR #372, opened 2024-03-28
Fix vcs segments

### LINKED ISSUE #371: VCS segments not working for non-git repositories nested/down the path
The current state of the vcs segments always only checks until the first match is found (which is git in all the vcs segments). This leads to showing only data for the nearest git repository up the path.

**An example**

`PWD` is `/home/exampleuser/projects/hg-test`

| Normal folder | Git repo             | Normal folder | hg/mercurial repo |
|--------------------|----------------------|--------------------|-------------------------|
| /home             |  /exampleuser   |  /projects        | /hg-test                 |

tmux-powerline will now show (e.g. for `vcs_branch`, `vcs_compare`, `vcs_others` and the others) the data for the git repository but not for the hg one.

xx4h:
* add lib/vcs_helper.sh containing logic to return the closest vcs repository and its type (git, hg, svn)
* adapt all vcs_* segments to use logic from new lib
* add config vars to most of the vcs segments

fixes #371

### DIFF
--- a/segments/vcs_revision.sh
+++ b/segments/vcs_revision.sh
@@ -3,41 +3,43 @@
 
 # Source lib to get the function get_tmux_pwd
 source "${TMUX_POWERLINE_DIR_LIB}/tmux_adapter.sh"
+source "${TMUX_POWERLINE_DIR_LIB}/vcs_helper.sh"
 
 run_segment() {
+	{ read vcs_type; read root_path; } < <(get_vcs_type_and_root_path)
 	tmux_path=$(get_tmux_cwd)
 	cd "$tmux_path"
 
-	stats=""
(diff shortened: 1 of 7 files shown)


segments/vcs_revision.sh
```diff
@@ -3,41 +3,43 @@
 
 # Source lib to get the function get_tmux_pwd
 source "${TMUX_POWERLINE_DIR_LIB}/tmux_adapter.sh"
+source "${TMUX_POWERLINE_DIR_LIB}/vcs_helper.sh"
 
 run_segment() {
+	{ read vcs_type; read root_path; } < <(get_vcs_type_and_root_path)
```
erikw, line 9: Cool, this construct is new to me

erikw (approved):
Good idea to refactor to lib

erikw:
Thank you, looks good!

PR #380, opened 2024-03-30
Add dual line status bar support

### LINKED ISSUE #379: Add dual-line status bar support
As i personally like to use a lot of segments and always have a fair amount of windows open, often it gets to the limits of a one-line status bar. So i added dual-line status bar support to tmux-powerline, already using it for a couple of days
now and not seeing any issues.

Even though this is expected to be working on Mac/BSD as well, testing on Mac/BSD is still to be done, i'll test this the coming days, but would also be happy if anyone wants to give it a try.

xx4h:
fixes #379


main.tmux
```diff
@@ -21,8 +21,15 @@ tmux set-option -g message-style "$TMUX_POWERLINE_STATUS_STYLE"
 tmux set-option -g status-left-length $TMUX_POWERLINE_STATUS_LEFT_LENGTH
 tmux set-option -g status-right-length $TMUX_POWERLINE_STATUS_RIGHT_LENGTH
 
-tmux set-option -g status-left "#(${TMUX_POWERLINE_DIR_HOME}/powerline.sh left)"
-tmux set-option -g status-right "#(${TMUX_POWERLINE_DIR_HOME}/powerline.sh right)"
+if [ "$TMUX_POWERLINE_STATUS_VISIBILITY" = "2" ]; then
+  tmux set-option -g status-format[$TMUX_POWERLINE_WINDOW_STATUS_LINE] "#[align=centre]#{W:#{E:window-status-format} ,#{E:window-status-current-format} }"
+  tmux set-option -g status-format[$TMUX_POWERLINE_OTHER_STATUS_LINE] "#[align=left]#(${TMUX_POWERLINE_DIR_HOME}/powerline.sh left)"
+  tmux set-option -ag status-format[$TMUX_POWERLINE_OTHER_STATUS_LINE] "#[align=right]#(${TMUX_POWERLINE_DIR_HOME}/powerline.sh right)"
+elif [ "$TMUX_POWERLINE_STATUS_VISIBILITY" = "on" ]; then
```
erikw, line 28: Could we move this to be the first check if the if-else, as this is the most common case? That will make the code easier to understand.

lib/config_file.sh
```diff
@@ -89,8 +89,11 @@ generate_default_config() {
 	# Overlay directory to look for segments. There you can put your own segments outside the repo. Fallback will still be the "segments" directory in the repo.
 	export TMUX_POWERLINE_DIR_USER_SEGMENTS="\${XDG_CONFIG_HOME:-\$HOME/.config}/tmux-powerline/segments"
 
-	# The initial visibility of the status bar. Can be {"on, off"}.
+	# The initial visibility of the status bar. Can be {"on", "off", "2"}. 2 will create two status lines.
 	export TMUX_POWERLINE_STATUS_VISIBILITY="${TMUX_POWERLINE_STATUS_VISIBILITY_DEFAULT}"
+	# In case of visibility = 2, where to display window status and where left/right status bars. 0 upper, 1 lower.
+	export TMUX_POWERLINE_WINDOW_STATUS_LINE=0
```
erikw, line 95: Are these two options mutually exclusive! If one is 0, must the other be 1?

In that case it’s bloated to expose 2 config names for this, then 1 is enough and derive the other one in the code from the first:).

erikw (changes requested):
Interesting!

Could you include a few screenshots in the PR how it looks?

Maybe mention the feature with one sentence in the README?

Oh and does the https://github.com/erikw/tmux-powerline/blob/main/mute_powerline.sh still work properly with 2 lines?

main.tmux
xx4h: Sure, done.

lib/config_file.sh
xx4h: You're right, the way it is implemented right now, they are more or less mutually exclusive.
So we have basically two options now:
1. Make it work, so that we can set both on 0, or both on 1. This would be pretty easy, but right now i could not think of a use case for having a dual-line status bar managed by tmux-powerline, but only using one the lines. 
3. Remove the `$TMUX_POWERLINE_OTHER_STATUS_LINE`

xx4h:
> Maybe mention the feature with one sentence in the README?
Good idea, yes, will do.

> Oh and does the https://github.com/erikw/tmux-powerline/blob/main/mute_powerline.sh still work properly with 2 lines?
yes, just tested, `{left, right, window-current-format, window-format}` can be individually muted and unmuted.

lib/config_file.sh
xx4h: I've decided for option 2, as there is no real scenario here right now and if it comes up, we'll implement it properly.

erikw: Since this is basically a Boolean configuration, I think it would be easiest for the user to have. Single confvar 

```bash
TMUX_POWERLINE_WINDOW_STATUS_SEPARATE=true/false
```

and letting it be `false` by default to keep things simple and opt-in.

erikw:
Great, thank you for testing the muting scripts too!

config/defaults.sh
```diff
@@ -5,6 +5,7 @@ export TMUX_POWERLINE_PATCHED_FONT_IN_USE_DEFAULT="true"
 export TMUX_POWERLINE_THEME_DEFAULT="default"
 
 export TMUX_POWERLINE_STATUS_VISIBILITY_DEFAULT="on"
+export TMUX_POWERLINE_WINDOW_STATUS_LINE=0
```
erikw, line 8: Good to add default value!

lib/config_file.sh
```diff
@@ -89,8 +89,11 @@ generate_default_config() {
 	# Overlay directory to look for segments. There you can put your own segments outside the repo. Fallback will still be the "segments" directory in the repo.
 	export TMUX_POWERLINE_DIR_USER_SEGMENTS="\${XDG_CONFIG_HOME:-\$HOME/.config}/tmux-powerline/segments"
 
-	# The initial visibility of the status bar. Can be {"on, off"}.
+	# The initial visibility of the status bar. Can be {"on", "off", "2"}. 2 will create two status lines.
```
erikw, line 92: Could clarify a bit more specifically 

```suggestion
	# The initial visibility of the status bar. Can be {"on", "off", "2"}. 2 will create two status lines: one for the window list and one with status bar segments. 
```

erikw (approved):
LGTM. Just one suggestion. Merge yourself when you deem it’s ready ☺️. Thank you for the cool feature!

xx4h:
Did some more tests, now also with macOS. Looking all good.

PR #394, opened 2024-04-07
Update air.sh

shyce:
Fix issue with posix -z flag


shyce:
image

While testing some emoji -> codepoint changes, noticed that the script wasn't running correctly

xx4h:
Thank you!

PR #395, opened 2024-04-07
vcs*: cleanup

xx4h:
* cleanup to comply with shellcheck
* better defaults, remove not needed __process_settings


erikw:
That’s a great initiative, to make the code comply with shellcheck!

Then at some point this workflow could be enabled again:

https://github.com/erikw/tmux-powerline/blob/main/.github/workflows/linter.yml.disabled

PR #397, opened 2024-04-07
Fix issue with air quality rendering a white background.

shyce:
This pull request fixes the default and bubble theme so that including `air` works properly by default.

- `air_color.sh` now exports TMUX_POWERLINE_SEG_AIR_COLOR
- Themes now source `air_color.sh`. I think there should be a standard interface for segments to provide suggested colors in the future but this works fine for now.
- Fixed up some extra spaces after comments.


xx4h:
Hi @shyce, the way i see it, `air_color.sh` should not be a segment at all. Overriding the color for a specific symbol of a segment should be done directly in the segment.
Something like we did here:
https://github.com/erikw/tmux-powerline/blob/fd480c2d6ae5e24a18589c4ef8e3685bc576e186/segments/vcs_branch.sh#L9
and then use it like this:
https://github.com/erikw/tmux-powerline/blob/fd480c2d6ae5e24a18589c4ef8e3685bc576e186/segments/vcs_branch.sh#L68

What do you think? Do you want to go for this refactor?

shyce:
Yeah let me take a look at this tonight after a bit of a break from the pc. I didn't like it either, just didn't want to commit a big change as a new contributor. Thanks for the feedback!

xx4h:
> Yeah let me take a look at this tonight after a bit of a break from the pc. I didn't like it either, just didn't want to commit a big change as a new contributor. Thanks for the feedback!

Cool, appreciate it! Let me know if you have any questions.

shyce:
I gave this a shot last night and it looks like it will need a bit more work. No modules are using #[bg=] currently because it doesn't color the powerline symbols as well, since it allows you to have multiple background colors in one echo. I'm still thinking about the best way to do this while adhering to the previous theme api.

I didn't want to change the way #[bg=] works because it might break peoples custom themes.

xx4h:
Oh, i guess i got i wrong yesterday. So it is not about the color for the symbol, it's about the color of the whole segment, right? And based on the air-quality the color of the segment should change?

segments/air_color.sh
```diff
@@ -5,4 +5,4 @@ if [ -n "$air_temp_file" ] && [ -f "$air_temp_file" ]; then
     TMUX_POWERLINE_SEG_AIR_COLOR=$(awk '{print $NF}' "$air_temp_file")
 fi
 
-TMUX_POWERLINE_SEG_AIR_COLOR="${TMUX_POWERLINE_SEG_AIR_COLOR:-'37'}"
+export TMUX_POWERLINE_SEG_AIR_COLOR="${TMUX_POWERLINE_SEG_AIR_COLOR:-'37'}"
```
xx4h, line 8:
```suggestion
echo "${TMUX_POWERLINE_SEG_AIR_COLOR:-'37'}"
```

themes/bubble.sh
```diff
@@ -60,7 +60,7 @@ TMUX_POWERLINE_SEPARATOR_THIN="|"
 # See Color formatting section below for details on what colors can be used here.
 TMUX_POWERLINE_DEFAULT_BACKGROUND_COLOR=${TMUX_POWERLINE_DEFAULT_BACKGROUND_COLOR:-$thm_bg}
 TMUX_POWERLINE_DEFAULT_FOREGROUND_COLOR=${TMUX_POWERLINE_DEFAULT_FOREGROUND_COLOR:-$thm_fg}
-TMUX_POWERLINE_SEG_AIR_COLOR=$("${TMUX_POWERLINE_DIR_HOME}/segments/air_color.sh")
+source ${TMUX_POWERLINE_DIR_HOME}/segments/air_color.sh
```
xx4h, line 63:
```suggestion
TMUX_POWERLINE_SEG_AIR_COLOR=$("${TMUX_POWERLINE_DIR_HOME}/segments/air_color.sh")
```

themes/default.sh
```diff
@@ -16,7 +16,7 @@ fi
 # See Color formatting section below for details on what colors can be used here.
 TMUX_POWERLINE_DEFAULT_BACKGROUND_COLOR=${TMUX_POWERLINE_DEFAULT_BACKGROUND_COLOR:-'235'}
 TMUX_POWERLINE_DEFAULT_FOREGROUND_COLOR=${TMUX_POWERLINE_DEFAULT_FOREGROUND_COLOR:-'255'}
-TMUX_POWERLINE_SEG_AIR_COLOR=$("${TMUX_POWERLINE_DIR_HOME}/segments/air_color.sh")
+source ${TMUX_POWERLINE_DIR_HOME}/segments/air_color.sh
```
xx4h, line 19:
```suggestion
TMUX_POWERLINE_SEG_AIR_COLOR=$("${TMUX_POWERLINE_DIR_HOME}/segments/air_color.sh")
```

themes/bubble.sh
```diff
@@ -129,13 +129,13 @@ if [ -z $TMUX_POWERLINE_LEFT_STATUS_SEGMENTS ]; then
...
 		"hostname $eggplant $thm_bg" \
-		# "ifstat 30 255" \
+		#"ifstat 30 255" \
 		#"ifstat_sys 30 255" \
 		"lan_ip $sky_blue $thm_bg ${TMUX_POWERLINE_SEPARATOR_RIGHT_THIN}" \
 		"wan_ip $sky_blue $thm_bg" \
 		"vcs_branch $thm_gray" \
-        # "air ${TMUX_POWERLINE_SEG_AIR_COLOR} $thm_bg" \
-		# "vcs_compare 60 255" \
+    #"air ${TMUX_POWERLINE_SEG_AIR_COLOR} $thm_bg" \
```
xx4h, line 137: Looks like here is still something wrong with the tabs?

xx4h (changes requested):
In this case i'd prefer not to source the segment in the theme, but rather fix the `air_color.sh` so it will echo back the color, so it can be used as expected in the theme files. I agree that this is still not a very good solution, but i guess it makes more sense to consider those kind of requirements (e.g. control the background color for the segment from within the segment itself) for a more general refactor.

shyce:
Yeah the air quality index changes the whole segment color based on the quality of the air I guess. I tried the echo approach instead of sourcing and for some reason, the output would not concat to the string. No idea why.

shyce:
I just tried again, and I guess I made a typo because it did work this time. Let me update it.

shyce:
Stand by, I guess my editor tab stops do not match the project.

shyce:
That's much cleaner, thanks. I'll try again at my dynamic approach that lets each segment setup their own colors in a later PR.

segments/air_color.sh
```diff
@@ -6,3 +6,4 @@ if [ -n "$air_temp_file" ] && [ -f "$air_temp_file" ]; then
 fi
 
 TMUX_POWERLINE_SEG_AIR_COLOR="${TMUX_POWERLINE_SEG_AIR_COLOR:-'37'}"
+echo $TMUX_POWERLINE_SEG_AIR_COLOR
```
xx4h, line 9:
```suggestion
echo "${TMUX_POWERLINE_SEG_AIR_COLOR:-'37'}"
```

xx4h: Re-assign unnecessary

PR #430, opened 2024-05-30
Fix segment wan_ip

matrixzj:
302 (redirect) response from Akamai. Fix with option `-L` follow redirect.

### DIFF
--- a/segments/wan_ip.sh
+++ b/segments/wan_ip.sh
@@ -39,7 +39,7 @@ run_segment() {
 	fi
 
 	if [ -z "$wan_ip" ]; then
-		if wan_ip=$(curl --max-time 2 -s -L https://whatismyip.akamai.com/); then
+		if wan_ip=$(curl --max-time 2 -s https://whatismyip.akamai.com/); then
 			echo "${wan_ip}" >"$tmp_file"
 		elif [ -f "${tmp_file}" ]; then
 			wan_ip=$(cat "$tmp_file")
--- a/segments/wan_ip.sh
+++ b/segments/wan_ip.sh
@@ -39,7 +39,7 @@ run_segment() {
 	fi
 
 	if [ -z "$wan_ip" ]; then
-		if wan_ip=$(curl --max-time 2 -s -L http://whatismyip.akamai.com/); then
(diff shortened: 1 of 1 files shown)


segments/wan_ip.sh
```diff
@@ -39,7 +39,7 @@ run_segment() {
 	fi
 
 	if [ -z "$wan_ip" ]; then
-		if wan_ip=$(curl --max-time 2 -s http://whatismyip.akamai.com/); then
+		if wan_ip=$(curl --max-time 2 -s -L http://whatismyip.akamai.com/); then
```
xx4h, line 42:
```suggestion
		if wan_ip=$(curl --max-time 2 -s https://whatismyip.akamai.com/); then
```

xx4h (changes requested):
Thanks for reporting, i suggest to directly use the new location (https) so we don't have two requests each run.

segments/wan_ip.sh
```diff
@@ -39,7 +39,7 @@ run_segment() {
 	fi
 
 	if [ -z "$wan_ip" ]; then
-		if wan_ip=$(curl --max-time 2 -s http://whatismyip.akamai.com/); then
+		if wan_ip=$(curl --max-time 2 -s -L https://whatismyip.akamai.com/); then
```
xx4h, line 42: I guess it was not clear what i meant. The reason why you get a 302, is the redirect from http to https. So we could either use `-L` to follow the redirect and keep the initial `http`, or we can just directly use `https` in which case we won't need the `-L` anymore.

```suggestion
		if wan_ip=$(curl --max-time 2 -s https://whatismyip.akamai.com/); then
```

matrixzj: Uh, I have got your points. But I think keeping `-L` will be helpful if akamai redirect it later.

xx4h: I don't expect any other redirects, as this is a special case due to switching to `https`-only.
If there will be more changes on akamai upstream, we'll address them correctly, but blindly following redirects will waste resources, obfuscate the usage of the wrong url (like you found out) and even introduces a security risk.

xx4h:
:+1: Thank you again for reporting!

PR #433, opened 2024-06-29
remove colour text from branch code

deltreey:
I was unable to get the `vcs` code to display properly, and I tracked the problem to this random text being inserted into the prompt.


deltreey (commented):
hmm, nevermind.  I'm an idiot

Issue #438, opened 2024-09-10
Bring back cute mode with hearts for battery segment


xx4h:
After reviewing the in #435 mentioned commit (https://github.com/erikw/tmux-powerline/commit/cab07470a7467068b6439a0cdd522be8c02e01fb), it looks like all the changes were intentional.

Before that commit the two modes configured through `TMUX_POWERLINE_SEG_BATTERY_TYPE`:
- `percentage`
- `cute`
were either displaying the current battery status in percentage, or by a number of hearts (`cute`).

The commit you mentioned has changed basically changed the looks of `cute` from hearts to a battery symbol.
So in general, nothing "broke", there's now just a new design for `cute`.

But to bring back the hearts, we should extend the possible values for `TMUX_POWERLINE_SEG_BATTERY_TYPE` by something like `battery` or `symbol`, to then have:
- `percentage` (showing battery status in percentage)
- `symbol` (showing battery status with battery symbol)
- `cute` (showing battery status with as many hearts as defined in `TMUX_POWERLINE_SEG_BATTERY_NUM_HEARTS` )

xx4h:
Fixed with #447

PR #447, opened 2024-10-07
Add option to display battery status as hearts in battery.sh

nikunjmathur08:
#438 

This PR restores the "hearts" display for the `cute` battery mode, as mentioned in issue #438. It also introduces a new `symbol` option to display the battery status with a battery symbol.


xx4h:
Thanks a lot for your work! 

You're using `BATTERY_FULL` and `BATTERY_EMPTY` inside of `__generate_hearts()` which will just use the battery symbols and not the hearts. I suggest we bring back `HEART_FULL` and `HEART_EMPTY`, removed by this commit: https://github.com/erikw/tmux-powerline/commit/cab07470a7467068b6439a0cdd522be8c02e01fb

Besides that, I just have 3 small remarks:
1. Add the missing `;;` mentioned by the linter
2. Change `TMUX_POWERLINE_SEG_BATTERY_NUM_BATTERIES_DEFAULT` to `TMUX_POWERLINE_SEG_BATTERY_NUM_HEARTS_DEFAULT` in 
https://github.com/erikw/tmux-powerline/blob/ba893c98eedc3311cc4dca4a21387c59ec4ad5dd/segments/battery.sh#L5 and https://github.com/erikw/tmux-powerline/blob/ba893c98eedc3311cc4dca4a21387c59ec4ad5dd/segments/battery.sh#L18 so it matches the other variable and the use case.
3. In https://github.com/erikw/tmux-powerline/blob/ba893c98eedc3311cc4dca4a21387c59ec4ad5dd/segments/battery.sh#L15 add the new `hearts` option to the doc line?

nikunjmathur08:
Made the following changes:

- Reintroduced HEART_FULL and HEART_EMPTY variables for displaying hearts.
- Renamed TMUX_POWERLINE_SEG_BATTERY_NUM_BATTERIES_DEFAULT to TMUX_POWERLINE_SEG_BATTERY_NUM_HEARTS_DEFAULT.
- Added the 'hearts' option to the TMUX_POWERLINE_SEG_BATTERY_TYPE documentation.
- Fixed the case statement by adding the missing ;; after the "hearts" case.

Fixes #438

nikunjmathur08:
I've also removed the extra 7th line.

segments/battery.sh
```diff
@@ -39,6 +41,9 @@ run_segment() {
 	"cute")
 		output=$(__cutinate "$battery_status")
 		;;
+	"hearts")
+		output=$(__generate_hearts "$battery_status")
```
xx4h, line 45:
```suggestion
		output=$(__generate_hearts "${battery_status/* /}")
```

While testing, i found out, that `$battery_status` is already containing a battery symbol. This is unfortunately baked into all the sub functions directly and needs a bigger rework. For now see my suggestion above.

xx4h:
LGTM, tested on linux.

Thanks again @nikunjmathur08!

nikunjmathur08:
@xx4h can you please add this under Hacktoberfest?

PR #458, opened 2025-01-25
added hostnamectl support

Ryuuusuke:
hostnamectl has native systemd support, and some users may not have inetutils installed. like me :)

### DIFF
--- a/segments/hostname.sh
+++ b/segments/hostname.sh
@@ -3,24 +3,33 @@
 
 TMUX_POWERLINE_SEG_HOSTNAME_FORMAT="${TMUX_POWERLINE_SEG_HOSTNAME_FORMAT:-short}"
 
+# check if user has hostname command
+has_hostname() {
+  command -v $1 >/dev/null
+}
+
 generate_segmentrc() {
-	read -r -d '' rccontents <<EORC
+  read -r -d '' rccontents <<EORC
 # Use short or long format for the hostname. Can be {"short, long"}.
 export TMUX_POWERLINE_SEG_HOSTNAME_FORMAT="${TMUX_POWERLINE_SEG_HOSTNAME_FORMAT}"
 EORC
-	echo "$rccontents"
+  echo "$rccontents"
 }
 
 run_segment() {
-	local opts=""
-	if [ "$TMUX_POWERLINE_SEG_HOSTNAME_FORMAT" == "short" ]; then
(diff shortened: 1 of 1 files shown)


segments/hostname.sh
```diff
@@ -3,24 +3,33 @@
 
 TMUX_POWERLINE_SEG_HOSTNAME_FORMAT="${TMUX_POWERLINE_SEG_HOSTNAME_FORMAT:-short}"
 
+# check if user has hostname command
+has_hostname() {
```
erikw, line 7: This method is actaully more general, it checks any program given in the arguments. Thus it could be named `command_exists` or something.

segments/hostname.sh
```diff
@@ -3,24 +3,33 @@
...
+  command -v $1 >/dev/null
+}
+
 generate_segmentrc() {
-	read -r -d '' rccontents <<EORC
```
erikw, line 7: Could you please update the PR to not change the indentation. I think it was already correct (hard tabs and not spaces). The project should be consistent, and PRs becomes more easy to follow if they focus on functional changes :)

segments/hostname.sh
```diff
@@ -3,24 +3,33 @@
...
+      opts="--short"
+    fi
+  fi
 
-	hostname ${opts}
-	return 0
+  if has_hostname hostname; then
+    hostname ${opts}
+  else
+    hostnamectl hostname
```
erikw, line 32: I don't think we can assume that if `hostname(1)` is not in `$PATH`, that `hostnamectl` is gong to be be. We must check in the same way if this command exist.

erikw (changes requested):
Thank you for the PR, it's a good idea to check which command is avaialble. A few changes are needed just to the PR :)

segments/hostname.sh
```diff
@@ -3,6 +3,10 @@
 
 TMUX_POWERLINE_SEG_HOSTNAME_FORMAT="${TMUX_POWERLINE_SEG_HOSTNAME_FORMAT:-short}"
 
+command_exists() {
```
erikw, line 6: Thank you. Acctually I see that this function would be good to make available in all segments. Could you move it to here?

https://github.com/erikw/tmux-powerline/blob/main/lib/util.sh

segments/hostname.sh
erikw: This is still needed :)

PR #463, opened 2025-03-28
Feat: Add a base16 default theme

scaryrawr:
This adds a theme based on the default/bubble that uses named tmux colors (the base16 palette).

This allows for theme colors to be inherited from
the terminal/shell, which is useful for things like VS Code color integration as well as other terminal with built in themes.

It also reads from a TMUX_POWERLINE_BUBBLE_SEPARATORS environment variable to toggle angled vs rounded separators.

color-change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ segments/xkb_layout
 themes/*
 !themes/default.sh
 !themes/bubble.sh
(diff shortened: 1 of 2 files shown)


erikw:
Cool one! Thank you for including the gif :-)

PR #464, opened 2025-04-12
Added "auto" option for the weather section

jackdeye:
Added the option to put "auto" in the latitude and longitude section of the theme, and if set, will query an api to get your location with your ip address.
Implements caching and a backup api call if the first does not work.


erikw (commented):
Nice idea, I like it!

I checked out the branch but unfortuantely it did not work for me. I set both long/lat config to `auto`. 

I see in the `/tmp/.weather_location_cache` the following:

```
set -x
exec 2>/tmp/tmux-powerline.log
<code to debug>
set +x
```

Then I inserted these `set -x` debug [commands ](https://github.com/jackdeye/tmux-powerline?tab=readme-ov-file#debugging) around the call to the function `get_auto_location()` to see the execution of this function:


> ++ get_auto_location
> ++ local cache_file=/tmp/.weather_location_cache
> ++ local max_cache_age=86400
> ++ [[ -f /tmp/.weather_location_cache ]]
> ++ local location_data
> ++ apis=("https://ipapi.co/json" "https://ipinfo.io/json")
> ++ local -a apis
> ++ for api in "${apis[@]}"
> +++ curl --max-time 4 -s https://ipapi.co/json
> ++ location_data='{
>     "error": true,
>     "reason": "RateLimited",
>     "message": "Visit https://ipapi.co/ratelimited/ for details"
> }'
> ++ case "$api" in
> +++ echo '{
>     "error": true,
>     "reason": "RateLimited",
>     "message": "Visit https://ipapi.co/ratelimited/ for details"
> }'
> +++ jq -r .latitude
> ++ TMUX_POWERLINE_SEG_WEATHER_LAT=null
> +++ echo '{
>     "error": true,
>     "reason": "RateLimited",
>     "message": "Visit https://ipapi.co/ratelimited/ for details"
> }'
> +++ jq -r .longitude
> ++ TMUX_POWERLINE_SEG_WEATHER_LON=null
> ++ [[ -n null ]]
> ++ [[ -n null ]]
> +++ dirname /tmp/.weather_location_cache
> ++ mkdir -p /tmp
> ++ echo 'export TMUX_POWERLINE_SEG_WEATHER_LAT='\''null'\'''
> ++ echo 'export TMUX_POWERLINE_SEG_WEATHER_LON='\''null'\'''
> ++ return 0
> ++ set +x
> jq: error (at <stdin>:1): Cannot index number with string "properties"
> parse error: Invalid numeric literal at line 1, column 8
> jq: error (at <stdin>:1): Cannot index number with string "properties"
> parse error: Invalid numeric literal at line 1, column 8

Looks like ratelimiting. If I visit in my browser https://ipapi.co/json it works.

jackdeye:
Thank you for checking this out! 
I updated the temporary directory structure to follow that pattern. I'm hoping the rate-limiting issue results from the cached file being unable to be located on your machine. I also removed the bash array to make it more accessible to older versions of bash.

erikw (approved):
Thanks for the update. I tried it again and now it works for me, no more rate limiting :-)

Issue #465, opened 2025-04-17
The Weather Segment (segments/weather.sh) does not work on MacOS despite using to GNU grep with Perl regular expression enabled


onlyuuuuu:
**Desc**

I'm on my Mac M1 and I was trying to setup my TMUX powerline with my own configurations according to the documentation. Everything goes well until one the segments, which is the `weather.sh` segment does seem to loaded up, no outputting text, just a small gap with blank content.

Upon further investigation, I've noticed that I'm using an unsupported `grep` with no `-P` (Perl expression support). Which is `grep (BSD grep, GNU compatible) 2.6.0-FreeBSD`. This is expected and have been mentioned in the documentation.

The solution will be to set the variable `TMUX_POWERLINE_SEG_WEATHER_GREP=ggrep` in my `config.sh` file since I'm using Homebrew and have `ggrep` installed, which is the GNU Grep and supported the `-P` (Perl expression support). However, the settings never get applied due to the implementation never applies the value of the variable and used it.

**Proposals**

The fix to this should be applied on line `97` from
```
grep=$TMUX_POWERLINE_SEG_WEATHER_GREP_DEFAULT
```
to
```
grep=$TMUX_POWERLINE_SEG_WEATHER_GREP
```

**Other Proposals/Issues**

On debugging and setting up the Weather segment. I've also found some other minor problems, such as:

1. The desc/docs on the segments has some outdated and misleading info. For ex:
```
# The data provider to use. Currently only "yahoo" is supported.
export TMUX_POWERLINE_SEG_WEATHER_DATA_PROVIDER="${TMUX_POWERLINE_SEG_WEATHER_DATA_PROVIDER_DEFAULT}"
```
Should be
```
# The data provider to use. Currently only "yrno" is supported.
export TMUX_POWERLINE_SEG_WEATHER_DATA_PROVIDER="${TMUX_POWERLINE_SEG_WEATHER_DATA_PROVIDER_DEFAULT}"
```
2. There are MacOS users, like me, uses tools from Homebrew GNU CoreUtils instead of the shitty FreeBDS stuff. So the `stat` command on line `83`
```
if shell_is_osx || shell_is_bsd; then
  last_update=$(stat -f "%m" "${tmp_file}")
```
doesn't work since we've sourced the CoreUtils stuff over the FreeBDS stuff (by writting it at the beginning of `$PATH`). So I have to do a workaround and added an entry in `config.sh` to something like this
```
[...]

PR #468, opened 2025-04-19
Modified the Weather segment script to be less complicated

onlyuuuuu:
Original Problem: https://github.com/erikw/tmux-powerline/issues/465

The purpose of this PR:
- Remove unnecessary complications, such as the usages of `stat` and `grep` command based on specific OS.
- Fix a bug when the location are set to `auto` only 1 API is called instead of 2, since `"null"` being handled as a legitimate value. Hence, if one of the API is rate limited, while the other one works fine. The real, actual coordinates never got fetched, which leads to weather API error due to lat and lon are set to `"null"`, as literal strings.
- Adding default values to lat and lon
- Allow tweaking the polling rate frequency of automatic location update (only works when both lat and lon settings are set to `auto`)

New script has been tested on Linux Ubuntu `amd64/aarch64` and MacOS `arm64`

### DIFF
--- a/segments/weather.sh
+++ b/segments/weather.sh
@@ -3,51 +3,64 @@
 # To configure your location, set TMUX_POWERLINE_SEG_WEATHER_LOCATION in the tmux-powerline config file.
 
 TMUX_POWERLINE_SEG_WEATHER_DATA_PROVIDER_DEFAULT="yrno"
-TMUX_POWERLINE_SEG_WEATHER_JSON_DEFAULT="jq"
 TMUX_POWERLINE_SEG_WEATHER_UNIT_DEFAULT="c"
 TMUX_POWERLINE_SEG_WEATHER_UPDATE_PERIOD_DEFAULT="600"
-
-if shell_is_bsd && [ -f /user/local/bin/grep ]; then
-	TMUX_POWERLINE_SEG_WEATHER_GREP_DEFAULT="/usr/local/bin/grep"
-else
-	TMUX_POWERLINE_SEG_WEATHER_GREP_DEFAULT="grep"
-fi
+TMUX_POWERLINE_SEG_WEATHER_LOCATION_UPDATE_PERIOD_DEFAULT="86400" # 24 hours
+TMUX_POWERLINE_SEG_WEATHER_LAT_DEFAULT="auto"
+TMUX_POWERLINE_SEG_WEATHER_LON_DEFAULT="auto"
 
 generate_segmentrc() {
 	read -r -d '' rccontents <<EORC
-# The data provider to use. Currently only "yahoo" is supported.
+# The data provider to use. Currently only "yrno" is supported.
 export TMUX_POWERLINE_SEG_WEATHER_DATA_PROVIDER="${TMUX_POWERLINE_SEG_WEATHER_DATA_PROVIDER_DEFAULT}"
 # What unit to use. Can be any of {c,f,k}.
 export TMUX_POWERLINE_SEG_WEATHER_UNIT="${TMUX_POWERLINE_SEG_WEATHER_UNIT_DEFAULT}"
 # How often to update the weather in seconds.
 export TMUX_POWERLINE_SEG_WEATHER_UPDATE_PERIOD="${TMUX_POWERLINE_SEG_WEATHER_UPDATE_PERIOD_DEFAULT}"
-# Name of GNU grep binary if in PATH, or path to it.
-export TMUX_POWERLINE_SEG_WEATHER_GREP="${TMUX_POWERLINE_SEG_WEATHER_GREP_DEFAULT}"
-# Location of the JSON parser, jq
-export TMUX_POWERLINE_SEG_WEATHER_JSON="${TMUX_POWERLINE_SEG_WEATHER_JSON_DEFAULT}"
+# How often to update the weather location in seconds (this is only used when latitude and longitude settings are set to "auto")
+export TMUX_POWERLINE_SEG_WEATHER_LOCATION_UPDATE_PERIOD="${TMUX_POWERLINE_SEG_WEATHER_LOCATION_UPDATE_PERIOD_DEFAULT}"
 # Your location
 # Latitude and Longtitude for use with yr.no
-# Set both to "auto" to detect automatically based on your IP address
-TMUX_POWERLINE_SEG_WEATHER_LAT=""
-TMUX_POWERLINE_SEG_WEATHER_LON=""
+# Set both to "auto" to detect automatically based on your IP address, or set them manually
+export TMUX_POWERLINE_SEG_WEATHER_LAT="${TMUX_POWERLINE_SEG_WEATHER_LAT_DEFAULT}"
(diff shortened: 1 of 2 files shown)


segments/weather.sh
```diff
@@ -118,8 +114,14 @@ __yrno() {
...
 		condition_symbol=$(__get_yrno_condition_symbol "$condition")
-		echo "${condition_symbol} ${degree}°$(echo "$TMUX_POWERLINE_SEG_WEATHER_UNIT" | tr '[:lower:]' '[:upper:]')" | tee "${tmp_file}"
+		# Write the <content @ date>, separated by 2 spaces and @, so we can fetch it later on without having to call 'stat'
+		echo "${condition_symbol} ${degree}°$(echo "$TMUX_POWERLINE_SEG_WEATHER_UNIT" | tr '[:lower:]' '[:upper:]')@$(date +%s)" > "$cache_weather_file"
+		__read_file_content "$cache_weather_file"
+		return
 	fi
+
+	__read_file_content "$cache_weather_file"
+	return
```
erikw, line 124: aaaand  just return some error value I guess here. because if we get here it means we could not retrieve the weather for some reason or the other.

```suggestion
	return 1
```

segments/weather.sh
```diff
@@ -118,8 +114,14 @@ __yrno() {
 		fi
 		# condition_symbol=$(__get_yrno_condition_symbol "$condition" "$sunrise" "$sunset")
 		condition_symbol=$(__get_yrno_condition_symbol "$condition")
-		echo "${condition_symbol} ${degree}°$(echo "$TMUX_POWERLINE_SEG_WEATHER_UNIT" | tr '[:lower:]' '[:upper:]')" | tee "${tmp_file}"
+		# Write the <content @ date>, separated by 2 spaces and @, so we can fetch it later on without having to call 'stat'
+		echo "${condition_symbol} ${degree}°$(echo "$TMUX_POWERLINE_SEG_WEATHER_UNIT" | tr '[:lower:]' '[:upper:]')@$(date +%s)" > "$cache_weather_file"
+		__read_file_content "$cache_weather_file"
+		return
```
erikw, line 120: Though we should move both reading and writing of cache to be done inside `run_segment()` to make it general for any weather provider. Right now only reading is done in the `run_segment()`, but we should do the writing there as well.

See the comment with pseudo code in `run_segment()` 😃 


Thus here we should just echo out the final result of this funciton `__yrno()` i.e.

```suggestion
		echo "${condition_symbol} ${degree}°$(echo "$TMUX_POWERLINE_SEG_WEATHER_UNIT" | tr '[:lower:]' '[:upper:]')@$(date +%s)"
		return 0
```

segments/weather.sh
```diff
@@ -177,23 +179,38 @@ __get_yrno_condition_symbol() {
 	esac
 }
 
-__read_tmp_file() {
-	if [ ! -f "$tmp_file" ]; then
+__read_file_split() {
```
erikw, line 182: Nice refactor, now this functinality can be reused.

Unfortuantely in bash we dont' have named arguments for functions with the benefit of documenting the function interface. What I like to do instead is to start each function by declaring local variables with explaining names as a way to document the interface, and subsequently make the function easier to understand with clear variable names.

Thus I would suggest to add this 

```suggestion
__read_file_split() {
	local file="$1"
	local lineno="$2"
	local errmsg="$3"
	[...]
```

erikw (changes requested):
👏 we're making this segment, which is probably the most loved and used tmux-powerline segment, more beautiful and efficient than it has ever been. I'm very glad to see that the GNU Grep usages are now gone here. I created an issue for removing GNU Grep usage in other segment, and added a new guidelines doc to prevent future PRs from adding it back! (https://github.com/erikw/tmux-powerline/blob/main/CONTRIBUTING.md)


Just a few more steps and then we are there. Very well done and good initiative making sense out of this old mess :-)

onlyuuuuu:
Cool, will get on that once I have the bandwidth

onlyuuuuu:
I'm back on this and looking at some suggestions from @erikw

segments/weather.sh
```diff
@@ -60,52 +68,40 @@ __process_settings() {
...
-			__read_tmp_file
-		fi
-	fi
 
 	if [ -z "$degree" ]; then
+		# There's a chance that you will get rate limited or both location APIs are not working
+		# Then long and lat will be "null", as literal string
+		if [ -z "$TMUX_POWERLINE_SEG_WEATHER_LAT" ] || [ -z "$TMUX_POWERLINE_SEG_WEATHER_LON" ] || [ "$TMUX_POWERLINE_SEG_WEATHER_LAT" == null ] || [ "$TMUX_POWERLINE_SEG_WEATHER_LON" == null ]; then
+			__read_file_content "$cache_weather_file"
+			return
```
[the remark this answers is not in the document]
onlyuuuuu: You will get both as "null" if the auto location dectection API is rate limited or it basically unreachable. So I'll be updating this to "Err: Unable to auto-detect your location" and returning 1. However, I have already done a check for "null" value in get_auto_location(), to skip the "null" values, so the last 2 or conditions will be removed too as they are now redundant.

segments/weather.sh
```diff
@@ -118,8 +114,14 @@ __yrno() {
 		fi
 		# condition_symbol=$(__get_yrno_condition_symbol "$condition" "$sunrise" "$sunset")
 		condition_symbol=$(__get_yrno_condition_symbol "$condition")
-		echo "${condition_symbol} ${degree}°$(echo "$TMUX_POWERLINE_SEG_WEATHER_UNIT" | tr '[:lower:]' '[:upper:]')" | tee "${tmp_file}"
+		# Write the <content @ date>, separated by 2 spaces and @, so we can fetch it later on without having to call 'stat'
```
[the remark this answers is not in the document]
onlyuuuuu: Tks. Actually, I got a hard time setting IFS=' @ ' with spaces as it doesn't work. Will update this comment, just a @ character as a separator, no spaces.

segments/weather.sh
```diff
@@ -3,51 +3,64 @@
...
+		*)
+			# Just read the stale cache, default to empty/blank
+			# Better not overwriting the previous healthy content
+			weather=$(__read_file_content "$cache_weather_file")
+			;;
+		esac
+		echo "$weather" > "$cache_weather_file"
+	fi
+
+	echo "$weather"
```
erikw, line 63: Wondrful 👏 🥳 🎆 , now this is the only echo for all code paths. This is a fantastic simplication of this old messy segment. Well don @onlyuuuuu !

segments/weather.sh
```diff
@@ -60,66 +73,52 @@ __process_settings() {
...
 		if ! get_auto_location; then
 			exit 8
 		fi
-	elif [ -z "$TMUX_POWERLINE_SEG_WEATHER_LON" ] || [ -z "$TMUX_POWERLINE_SEG_WEATHER_LAT" ]; then
-		echo "No location defined." >&2
-		exit 8
 	fi
 }
 
+# An implementation of a weather provider, just need to echo the result, run_segment() will take care of the rest
```
erikw, line 86: Good idea to add a note. Thanks!

erikw (approved):
Nicely done mate! Big thanks to you 👏

PR #480, opened 2025-05-15
fix: get power supplies of type battery

### LINKED ISSUE #455: Questions about BAT0 or BAT1 in battery.sh
it doesn't exist on my lattop (LG, Ubuntu 24.04 and Ubuntu 20.04). My laptop only has CMB0. I replaced BAT0 in .sh with CMB0 to make it work properly.

xx4h:
* cleanup unused variables
* remove no longer needed directory test

fix #455

### DIFF
--- a/segments/battery.sh
+++ b/segments/battery.sh
@@ -104,18 +104,6 @@ __battery_macos() {
 __battery_linux() {
 	case "$SHELL_PLATFORM" in
 	"linux")
-		BATPATH=/sys/class/power_supply/BAT0
-		if [ ! -d $BATPATH ]; then
-			BATPATH=/sys/class/power_supply/BAT1
-		fi
-		BAT_FULL=$BATPATH/charge_full
-		if [ ! -r $BAT_FULL ]; then
-			BAT_FULL=$BATPATH/energy_full
-		fi
-		BAT_NOW=$BATPATH/charge_now
-		if [ ! -r $BAT_NOW ]; then
-			BAT_NOW=$BATPATH/energy_now
-		fi
 		__linux_get_bat
 		;;
 	"bsd")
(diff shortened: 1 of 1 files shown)


segments/battery.sh
```diff
@@ -158,28 +146,26 @@ __linux_get_bat() {
...
+		if [ -r "$full" ] && [ -r "$now" ]; then
+			local bf
+			local bn
+			bf=$(cat "$full")
+			bn=$(cat "$now")
+			total_full=$((total_full + bf))
+			total_now=$((total_now + bn))
 		fi
-	done
+	done <<<"$(grep -l "Battery" /sys/class/power_supply/*/type | sed -e 's,/type$,,')"
```
erikw, line 168: This looks like a better solution!

segments/battery.sh
```diff
@@ -158,28 +146,26 @@ __linux_get_bat() {
...
-				bn=$(cat "$now")
-				total_full=$((total_full + bf))
-				total_now=$((total_now + bn))
-			fi
+		if [ -r "$full" ] && [ -r "$now" ]; then
+			local bf
+			local bn
+			bf=$(cat "$full")
+			bn=$(cat "$now")
+			total_full=$((total_full + bf))
```
erikw, line 165: Good, summing up total battery is preserved.

erikw (approved):
I can't test the updated code myself at this moment, but the code changes looks good as far as I can see.

xx4h:
> I can't test the updated code myself at this moment, but the code changes looks good as far as I can see.

I should have mentioned, that i tested it.
Only thing i can think of right now that could mess things up would be batteries with wrong type, or non-batteries with type "Battery".

PR #484, opened 2025-07-04
feat(weather): use new error logging helper

xx4h:


xx4h:
depends on #483 (therefore duplicated changes)

PR #485, opened 2025-07-04
feat(weather): use new logging helper

xx4h:


xx4h:
depends on #483

segments/weather.sh
```diff
@@ -281,6 +284,10 @@ get_auto_location() {
         fi
     fi
 
-    echo "Could not detect location automatically" >&2
+    __err "Could not detect location automatically"
     return 1
 }
+
+__err() {
```
erikw, line 291: To avoid needing to make a error helper in every segment, then we don't gain much by having on in the utils, could we do like this?


When a segment is being processed, it looks like we can get the name of the current segment. Then `err` in the lib does not need the first argument, it can get it like this:

https://github.com/erikw/tmux-powerline/blob/a14267b48f26e6b44b1c378374b8b82e522209c3/lib/powerline.sh#L113

Maybe this `_process_scripts()` should actually export the current segment name in an environment variable that can be used like `TMUX_POWERLINE_CURRENT_SEGMENT` before this function calls `run_segment()` for the current segment.

erikw (changes requested):
See comment in the other PR

erikw (approved):
Looks much better now after change in the other PR!

Wait to rebase this on the other PR.

erikw (approved):
🚀

PR #490, opened 2025-07-28
feat: allow customizing powerline depending on client_width

cousine:
This commit sets `COLUMNS` to tmux's `client_width` to allow customizing segments using the `$COLUMNS` environment variable, eg of a theme utilizing the variable to detect available width:

```
  if [ $COLUMNS -le 147 ]; then
    TMUX_POWERLINE_RIGHT_STATUS_SEGMENTS=(
      "time 4 17 ${TMUX_POWERLINE_SEPARATOR_LEFT}"
      "battery 1 255"
    )
  elif [ $COLUMNS -le 201 ]; then
    TMUX_POWERLINE_RIGHT_STATUS_SEGMENTS=(
      "now_playing 13 17"
      "date 4 17"
      "time 4 17 ${TMUX_POWERLINE_SEPARATOR_LEFT_THIN}"
      "battery 1 255"
    )
  else
    TMUX_POWERLINE_RIGHT_STATUS_SEGMENTS=(
      "now_playing 13 17"
      "date 4 17"
      "time 4 17 ${TMUX_POWERLINE_SEPARATOR_LEFT_THIN}"
      "battery 1 255"
      #"utc_time 235 136 ${TMUX_POWERLINE_SEPARATOR_LEFT_THIN}" \
    )
```

This is especially useful in multi-display setups

### DIFF
--- a/segments/np_spotify_mac.script
+++ b/segments/np_spotify_mac.script
@@ -15,7 +15,7 @@ if process_list contains "Spotify" then
 			if player state is playing then
 				set now_playing to "► " & artist_name & " - " & track_name
 			else
-				set now_playing to "❙❙ " & artist_name & " - " & track_name
+				set now_playing to ""
 			end if
 			if length of now_playing is less than trim_length then
 				set now_playing_trim to now_playing


main.tmux
```diff
@@ -27,8 +27,8 @@ tmux set-option -g message-style "$TMUX_POWERLINE_STATUS_STYLE"
 tmux set-option -g status-left-length "$TMUX_POWERLINE_STATUS_LEFT_LENGTH"
 tmux set-option -g status-right-length "$TMUX_POWERLINE_STATUS_RIGHT_LENGTH"
 
-tmux set-option -g status-left "#(${TMUX_POWERLINE_DIR_HOME}/powerline.sh left)"
-tmux set-option -g status-right "#(${TMUX_POWERLINE_DIR_HOME}/powerline.sh right)"
+tmux set-option -g status-left "#(COLUMNS=#{client_width} ${TMUX_POWERLINE_DIR_HOME}/powerline.sh left)"
```
erikw, line 30: There's no need to do this. If a script needs to look up the `client_with`, or any other tmux variable, it can be done like this in the script:

```bash
cw=$(tmux display-message -p "#{client_width}")
```

Would this work for you?

If so, I would prefer this not to bloat the standard setup. After all, there are over 200 [tmux variables](https://man7.org/linux/man-pages/man1/tmux.1.html) and we can't add each here as an env var....

erikw (commented):
Where does the example code in the PR description live? You mentioned customization in segments, but this looks more like code that would live in a theme to me.

As I understand it, this is not a good way forward to start adding each tmux variable as an environment variable. See the PR comment below for instructions on how to still obtain the `client_width` for your multi-setup.

Let me know if I misunderstood the intention or problem :).

main.tmux
cousine: Thanks for taking a look! while this approach works, it has a side effect where the `cw` value will change when switching back and forth between sessions on different displays

cousine: I understand of course the bloat concern, but I would say COLUMNS is a standard variable that should be set anyway (unless there is a way to avoid the side effect I mention above) 🤔

xx4h: > I understand of course the bloat concern, but I would say COLUMNS is a standard variable that should be set anyway (unless there is a way to avoid the side effect I mention above) 🤔

Afaik `$COLUMNS` is only being set/available in interactive sessions. So that's why you'll see it in your normal terminal bash sessions (even if they are in a tmux window), but they are not available in the bash tmux itself is running e.g. for tmux-powerline.
This is also why i would not want to set `$COLUMNS` for those kind of non-interactive sessions.

> Thanks for taking a look! while this approach works, it has a side effect where the `cw` value will change when switching back and forth between sessions on different displays

This is most probably due to shared server session from tmux, you could try to spawn those separate tmux sessions with something like
```bash
tmux new-session -L tmux1
tmux new-session -L tmux2
tmux new-session -L tmux3
```
I expect this to resolve this issue.

erikw: Thanks @xx4h for clarifying how `COLUMNS` work and sessions!

xx4h: Just to avoid future confusion, cite from tmux man page:

> The tmux server manages clients, sessions, windows and panes.

So my wording

> This is most probably due to shared server session from tmux, ...

is not really correct and the word "session" may be confused with the actual tmux sessions "inside" of a tmux server. So i suggest you don't name them `$SESSION_NAME` but rather `$SERVER_NAME` or even better `$SERVER_SOCKET_NAME`.

PR #496, opened 2025-10-30
Memory used segment

iyegoroff:
A simpler version of "tmux-mem-cpu-load" plugin, which doesn't require dependencies & occupies less space. Memory usage data is identical to data from "tmux-mem-cpu-load", despite the [calculation](https://github.com/iyegoroff/tmux-powerline/blob/mem_used/lib/mem_used_helper.sh#L18) for Macos doesn't explicitly subtract "speculative_count" from "total_mem" like in [tmux-mem-cpu-load](https://github.com/thewtex/tmux-mem-cpu-load/blob/master/osx/memory.cc#L60-L63). It looks like a difference lies between "vm_stat" program used in this segment and native "host_statistics64" function from "tmux-mem-cpu-load" - "vm_stat" itself subtracts "speculative_count" from "free_count" while "host_statistics64" avoids that step.

Linux (left side - mem_used, right side - tmux-mem-cpu-load):

Macos:
Screenshot 2025-10-30 at 11 44 05

GB instead of MB - identical to data from "Activity Monitor" app on Macos:
Screenshot 2025-10-30 at 11 42 31

High memory load colors:
Screenshot 2025-10-30 at 11 45 14
Screenshot 2025-10-30 at 11 45 35

### DIFF
--- a/lib/mem_used_helper.sh
+++ b/lib/mem_used_helper.sh
@@ -31,7 +31,7 @@ __tp_mem_used_info() {
 		local s_reclaimable
 		local mem_used_bytes
 
-		meminfo=$(cat /proc/meminfo | tr '\n' ' ')
+		meminfo=$(tr '\n' ' ' < /proc/meminfo)
 		mem_total=$(echo "$meminfo" | sed -e 's/^MemTotal: *\([0-9]*\).*/\1/')
 		mem_total_bytes=$(echo "$mem_total * 1024" | bc -l)
 		mem_free=$(echo "$meminfo" | sed -e 's/.* MemFree: *\([0-9]*\).*/\1/')


segments/mem_used.sh
```diff
@@ -0,0 +1,50 @@
...
+__process_settings() {
+	if [ -z "$TMUX_POWERLINE_SEG_MEM_USED_ICON" ]; then
+		export TMUX_POWERLINE_SEG_MEM_USED_ICON="${TMUX_POWERLINE_SEG_MEM_USED_ICON_DEFAULT}"
+	fi
+	if [ -z "$TMUX_POWERLINE_SEG_MEM_USED_UNIT" ]; then
+		export TMUX_POWERLINE_SEG_MEM_USED_UNIT="${TMUX_POWERLINE_SEG_MEM_USED_UNIT_DEFAULT}"
+	fi
+};
+
+# source https://askubuntu.com/a/179949
```
erikw: Could you please add a line or two of documentation on how this is rounding? Maybe an example.

Would it make sense to let the `tp_mem_used_*bytes()` functions return the rounded result directly?

PR #510, opened 2026-05-12
fix(now_playing): remove hardcoded trim from macOS scripts, enhance Apple Music

### LINKED ISSUE #508: now_playing: text does not roll if it is set to the default 40
I noticed that if `TMUX_POWERLINE_SEG_NOW_PLAYING_MAX_LEN` is left at the default `40`,  the text will not roll, even if the trim method is set to roll. Dug into the code a bit and found that it was because the relevant AppleScripts first trim the length to 40 — so the text that comes through to the bash script is always limited to 40.

Was there a particular reason why the AppleScripts need to trim the text? If not, I can put up a PR to remove the trim in the AppleScripts and let the bash script handle trimming.

deanbear:
## Summary

Closes #508.

The macOS now_playing scripts (`np_apple_music.script`, `np_itunes.script`, `np_spotify_mac.script`) all contained a hardcoded 40-character trim that was applied before the output reached the bash layer. This meant `TMUX_POWERLINE_SEG_NOW_PLAYING_MAX_LEN` and `TRIM_METHOD=roll` had no effect for these players — the text was always pre-truncated to 40 chars and the roll logic never triggered.

The fix removes the trim from all three scripts and lets `now_playing.sh` handle length uniformly, consistent with how all other players work.

### Changes

- **`np_apple_music.script`**: Remove hardcoded trim. Add paused state support and ►/❙❙ play/pause icons to match `np_spotify_mac.script` behaviour.
- **`np_itunes.script`**: Remove hardcoded trim.
- **`np_spotify_mac.script`**: Remove hardcoded trim (play/pause icons were already present).

## Test plan

- [ ] Set `TRIM_METHOD=roll` and `MAX_LEN=40` — confirm text rolls when track name exceeds 40 chars
- [ ] Pause Apple Music — confirm track info remains visible with ❙❙ icon
- [ ] Play Apple Music — confirm ► icon appears
- [ ] Verify Spotify and iTunes behaviour unchanged (trim/roll now handled by bash)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/segments/np_apple_music.script
+++ b/segments/np_apple_music.script
@@ -1,5 +1,5 @@
 #!/usr/bin/env osascript
-# Returns the current playing song in iTunes for macOS.
+# Returns the current playing song in Apple Music for macOS.
 
 tell application "System Events"
 	set process_list to (name of every process)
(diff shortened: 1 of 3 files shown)


erikw (approved):
LGTM. I tried it out locally.

Issue #351, opened 2023-12-18
Screen scrolling on this on and duplicationg the status bar


Dominiquini:
As soon I start tmux, the screen start scrolling randomly (sometimes it takes several seconds to scroll 1 line, but other times it scroll several lines in 1 second), and the status bar end up duplicated! The prompt on the top also disappears (but the cursor remains and I was still able to type!)

![WindowsTerminal_dqMN4nmOiE](https://github.com/erikw/tmux-powerline/assets/1180808/0b7581f3-f5f9-4583-b88f-380959aa928e)
![WindowsTerminal_K5mKOw7jry](https://github.com/erikw/tmux-powerline/assets/1180808/7f767c41-3c97-4a51-8bc9-cfea137faa55)

I tested removing all plugins from my .tmux.conf, leaving only 'tpm' and this!
Tried also deleting my .bashrc to avoid any configuration that cam interfere!

I installed this plugin some days ago and everything appears to be working fine. Maybe some recent update that caused this bug!

OS: Ubuntu 22.04 (WSL2 under Windows 10 x64)
SHELL: GNU bash 5.1.16
TMUX: 3.2a

Is there anything that I can provide to help in this bug?

Thanks.

Dominiquini:
Doing more tests here, I found out that is the weather segment that is causing this issue. Disabling this segment on the default theme resolve this issue!

Thanks.

killbasa:
The issue for me was the fact that some of the weather icons are 2 characters in width rather than 1. Main clue was that my status bar was fine in the morning, but the above issue appeared later in the night.

Arcadi4:
Turning off the weather segment also worked for me.

xx4h:
Trying to reproduce this. From the screenshots my assumption is, that this was weather `cloudy` and @Dominiquini stated, that he's using Ubuntu 22.04 in WSL2.
I'm thinking if this is maybe WSL2 related, did you also see the bug in WLS2 (@killbasa  @Arcadi4)? If not, could you maybe give me some more information on what environment/terminal/etc. you're using?

killbasa:
> Trying to reproduce this. From the screenshots my assumption is, that this was weather `cloudy` and @Dominiquini stated, that he's using Ubuntu 22.04 in WSL2. I'm thinking if this is maybe WSL2 related, did you also see the bug in WLS2 (@killbasa @Arcadi4)? If not, could you maybe give me some more information on what environment/terminal/etc. you're using?

[...]

PR #511, opened 2026-05-14
fix(weather): add nerd-font icon set and improve emoji width handling

deanbear:
## Summary

- Add `TMUX_POWERLINE_SEG_WEATHER_ICON_SET` config option (`emoji` / `nerd-font`) to allow users to switch to Nerd Font weather icons for consistent terminal width rendering
- In emoji mode, add VS16 (U+FE0F) to EAW=Neutral emoji so tmux 3.4+ correctly calculates their width via `variation-selector-always-wide`
- `nerd-font` mode uses Nerd Font icons with guaranteed consistent width, independent of tmux version — recommended for users experiencing status bar alignment issues
- Improve emoji semantic mapping: distinguish light/normal/heavy rain, separate sleet from rain group, split snow by intensity
- Fix `heavysleetshowersandthunder_night` incorrectly grouped in showers_day instead of thunder
- Replace `clearsky_day` icon from ☀️ (VS16, unreliable width) to 🌞 (EAW=W, consistent width)
- Cache key includes icon set to avoid stale icons after switching

## Context

See detailed analysis in https://github.com/erikw/tmux-powerline/issues/351#issuecomment-4450317537

The root cause is a mismatch between terminal emoji display width and tmux's internal `wcwidth()` calculation. 

I'm not sure this PR is worth merging as-is. 

The real fix for the emoji width issue is upgrading tmux to 3.4+, which correctly handles VS16. This PR doesn't solve the root cause — it provides a `nerd-font` fallback for users on older tmux versions and improves the emoji mappings to work better with tmux 3.4+'s VS16 handling.

## Test plan

- [ ] Verify weather segment displays correctly with `emoji` icon set (default)
- [ ] Verify weather segment displays correctly with `nerd-font` icon set
- [ ] Verify switching icon set invalidates cache (no stale icons)
- [ ] Verify all 62 met.no weather conditions are covered in both icon sets
- [ ] Test on tmux < 3.4 with `nerd-font` to confirm no alignment issues

Ref: #351

### DIFF
--- a/segments/weather.sh
+++ b/segments/weather.sh
@@ -11,10 +11,9 @@ TMUX_POWERLINE_SEG_WEATHER_UPDATE_PERIOD_DEFAULT="600"
 TMUX_POWERLINE_SEG_WEATHER_LOCATION_UPDATE_PERIOD_DEFAULT="86400" # 24 hours
 TMUX_POWERLINE_SEG_WEATHER_LAT_DEFAULT="auto"
 TMUX_POWERLINE_SEG_WEATHER_LON_DEFAULT="auto"
+TMUX_POWERLINE_SEG_WEATHER_ICON_SET_DEFAULT="emoji"
 
-# Global cache file for weather data
-TMUX_POWERLINE_SEG_WEATHER_CACHE_FILE_WEATHER="${TMUX_POWERLINE_DIR_TEMPORARY}/weather_cache_data.txt"
-# Add: global cache file for auto-detected location (lat/lon)
+# Global cache file for auto-detected location (lat/lon)
 TMUX_POWERLINE_SEG_WEATHER_CACHE_FILE_LOCATION="${TMUX_POWERLINE_DIR_TEMPORARY}/weather_cache_location.txt"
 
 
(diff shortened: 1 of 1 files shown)


xx4h:
Thank you for your PR @deanbear

As we already have Nerd Fonts in the requirements for tmux-powerline
https://github.com/erikw/tmux-powerline/blob/6cddf8051823815db77fc00dcbb6bcca9afbdec2/README.md?plain=1#L101
we should just add the Nerd Font icons to `__get_yrno_condition_symbol` directly. @erikw, what's your take on this?

A reproducible test case would be very much appreciated, see my comment in [#351 ](https://github.com/erikw/tmux-powerline/issues/351#issuecomment-4458075008)

erikw:
Hey,

I don’t have laptop access right now so it’s hard to look at details. It’s a throughout analysis in the issue. 

* Do we have a reproducible test case to verify that this indeed was the root cause?


I think we should avoid making the configuration more complicated with a different emoji set. 

@xx4h ” we should just add the Nerd Font icons to __get_yrno_condition_symbol directly” sounds like an easier solution if this solves the problem (can’t reasonably look deeper at what that means with my phone at the moment).

erikw:
Superseeded by #514 and #515

PR #512, opened 2026-05-16
Add config options to pass --player and --ignore-player options to playerctl.

danyael031:
Added TMUX_POWERLINE_SEG_NOW_PLAYING_PLAYERCTL_OPT_PLAYER and TMUX_POWERLINE_SEG_NOW_PLAYING_PLAYERCTL_OPT_IGNORE_PLAYER options config options to let user pass --player and --ignore-player values to playerctl to control its behavior.


erikw (changes requested):
Looks reasonable to me. Could please fix the linter warnings just? :)

danyael031:
After taking a look to the linter issue (SC2086) I've decided to change the implementation, so instead of allowing the user to pass any arbitrary option, the implementation only contemplates --player and --ignore-player, since any other option might be suboptimal and break the formatting of playerctl's output.

erikw:
Thank you :)

Issue #266, opened 2018-06-25
tmux-powerline weather bug


JoeAsir:
## when I actived the weather in default.sh, the bar would continue be updating and never stop. I have no idea what's going on and how to fix it up.
![image](https://user-images.githubusercontent.com/14145765/41831355-195dfb2a-7879-11e8-8782-81484b7e6879.png)

wu-nerd:
Same problem here, it just happened, everything works fine before with the same configuration. 
Kill tmux or restart the computer changed nothing, delete  TMUX_POWERLINE_SEG_WEATHER_LOCATION variable it's back to normal.

tmux 2.7 
macOS 10.13.6 (17G65)

wu-nerd:
Found workaround in https://github.com/erikw/tmux-powerline/issues/242:
Comment https://github.com/erikw/tmux-powerline/blob/e214721694661448176580433fd563f82542545a/segments/weather.sh#L128 and set condition_symbol to a common character, like a space.

It works, but then i can not reproduce this problem. Weird.

JoeAsir:
Really works! Thx

PR #513, opened 2026-05-27
fix(segments): make weather and wan_ip fetches non-blocking; add weather icon style option

turtle11311:
## Problem

`weather.sh` and `wan_ip.sh` perform synchronous `curl` calls whenever their cache expires. During this window, `powerline.sh` blocks for up to **12 seconds** (weather: location API × 2 + forecast API) or **2 seconds** (wan_ip), causing tmux to display a `<'powerline.sh ...' not ready>` placeholder and the status bar to visually push terminal content upward.

This has been reported in:
- #351 — Screen scrolling & status bar duplication
- #266 — weather segment causes status bar to update continuously

The issue is intermittent because it only triggers when the cache expires (every 10 min for weather, every 15 min for wan_ip), making it hard to trace back to the segments.

## Root Cause

Both segments used the pattern:
```
cache fresh → return immediately
cache stale → synchronous curl → update cache → return
```

The stale branch blocks the entire `powerline.sh` process, preventing tmux from rendering the status bar in time.

A secondary issue in `weather.sh`: condition symbols using VS16 variation selectors (e.g. `☀️` = U+2600 + U+FE0F) cause tmux to miscount the status-bar width. The base character (U+2600) has East Asian Width: Neutral (1 cell), but VS16 makes the terminal render it as 2 cells. The 1-cell discrepancy causes the status bar to overflow and wrap, pushing terminal content up. This only triggers for specific weather conditions (clearsky, cloudy, thunderstorm, snow) which explains why the symptom appears intermittently.

### Non-blocking fetch (both segments)

The cache is now the sole data source for `run_segment()`. It always returns immediately — even stale data is better than blocking:

```
always return cached value (even if stale)
cache stale or missing → spawn background subshell to refresh
```

A lock file prevents concurrent background refreshes. On first run with no cache, the segment displays empty until the background fetch completes (a few seconds).

### Weather icon style option (`weather.sh` only)

Adds `TMUX_POWERLINE_SEG_WEATHER_ICON_STYLE` to let users opt in to width-safe icon rendering. The default preserves the original behaviour so no existing setup is affected:

| Value | Behaviour |
|---|---|
| `emoji` | Original emoji with VS16 preserved **(default)** |
| `emoji_fixed` | VS16 stripped — for terminals where tmux miscounts width |
| `nerdfonts` | Nerd Font PUA icons (fixed 1-cell width, no ambiguity) |
| `auto` | `nerdfonts` if `tp_patched_font_in_use()`, else `emoji` |

## Testing

Tested on Linux (Ghostty terminal, tmux 3.x) over several days. Verified:
- Status bar no longer pushes up during cache refresh cycles
- `time bash powerline.sh right` returns in ~0.15s regardless of cache state
- All four icon styles render correctly

### DIFF
--- a/segments/wan_ip.sh
+++ b/segments/wan_ip.sh
@@ -1,5 +1,7 @@
 # shellcheck shell=bash
 # Prints the WAN IP address. The result is cached and updated according to $update_period.
+#
+# Network fetch is done in a background process so tmux rendering is never blocked.
 TMUX_POWERLINE_SEG_WAN_IP_SYMBOL="${TMUX_POWERLINE_SEG_WAN_IP_SYMBOL:-ⓦ }"
 TMUX_POWERLINE_SEG_WAN_IP_SYMBOL_COLOUR="${TMUX_POWERLINE_SEG_WAN_IP_SYMBOL_COLOUR:-255}"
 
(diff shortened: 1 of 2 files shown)


turtle11311:
To clarify the scope of this PR relative to #511:

**This PR fixes two independent issues:**

### 1. Blocking network calls (not covered by #511)

The status bar push-up most users experience is actually caused by `run_segment()` blocking on `curl` when the cache expires — not (only) by emoji width. During the blocking window, tmux displays `<'powerline.sh ...' not ready>` and the layout shifts.

**Reproduction steps:**
```bash
# 1. Enable weather or wan_ip segment
# 2. Delete the cache to force a cold fetch
rm /tmp/tmux-powerline_${USER}/weather_cache_data.txt

# 3. Watch the next status-interval tick — you will see:
#    <'/path/to/powerline.sh right' not ready>
#    and the terminal content pushed up

# 4. Confirm the fetch is slow:
time bash ~/.tmux/plugins/tmux-powerline/powerline.sh right
# worst case: ~12s (location API x2 + forecast API, all with --max-time 4)
```

This PR changes `run_segment()` to always return the cached value immediately and refresh in the background, so `powerline.sh` consistently returns in ~0.15s regardless of cache state.

### 2. Nerd Font icon option

The `TMUX_POWERLINE_SEG_WEATHER_ICON_STYLE` option was added to address the emoji width issue independently of tmux version. Since Nerd Font is already a project requirement, this also aligns with @xx4h's suggestion to add Nerd Font icons directly to `__get_yrno_condition_symbol`. The default remains `emoji` to preserve existing behaviour.

There is overlap with #511 on the emoji width front — happy to coordinate or adjust scope based on maintainer feedback.

erikw:
Hello @turtle11311, thank you for looking into this. These are two different issues and two different fixes. It would be best to split this up into 2 PRs so they can be reviewed and tested separately. It will help future regression debugging as well.

turtle11311:
Closing in favour of two focused PRs as requested by @erikw:

- #514 — non-blocking fetch for `weather` and `wan_ip` segments (+ lock file staleness guard + stderr isolation)
- #515 — `TMUX_POWERLINE_SEG_WEATHER_ICON_STYLE` option for emoji width fix and Nerd Font icons

Both PRs incorporate the improvements discussed in the review comments on this PR.

PR #514, opened 2026-05-27
fix(segments): make weather and wan_ip fetches non-blocking

turtle11311:
## Problem

When the `weather` or `wan_ip` segment cache expires, `run_segment()` calls `curl` synchronously. During that window:

- `powerline.sh` blocks for up to **~12 s** (weather: location API × 2 + forecast API, each `--max-time 4`) or **~2 s** (wan_ip)
- tmux displays `<'powerline.sh …' not ready>` in the status bar and **pushes terminal content up**
- This repeats on every cache-expiry cycle (every 600 s for weather, 900 s for wan_ip)

Reported in #351 (status-bar duplication / push-up) and #266 (weather causing repeated updates). The root cause — synchronous network calls inside `run_segment()` — was not previously identified.

## Fix

`run_segment()` now always returns the cached value immediately (even if stale), so `powerline.sh` returns in **~0.15 s** regardless of cache state. When the cache is stale or missing, a background subshell refreshes it; the updated value appears on the next render cycle.

Additional hardening:
- Lock files older than the segment's maximum possible fetch time (10 s for `wan_ip`, 30 s for `weather`) are treated as abandoned (e.g. left by a prior `SIGKILL`) and removed automatically.
- `stderr` is redirected to `/dev/null` inside the background subshell to prevent `curl` errors from leaking into tmux's output context after the parent has already returned.

## Before / After

| Scenario | Before | After |
|---|---|---|
| `powerline.sh` return time on cache miss | up to ~12 s | ~0.15 s |
| `not ready` placeholder visible | every cache-expiry cycle | never |
| Terminal content pushed up | every cache-expiry cycle | never |

## Reproduction (before fix)

```bash
# 1. Enable weather or wan_ip segment
# 2. Force a cold cache
rm /tmp/tmux-powerline_${USER}/weather_cache_data.txt

# 3. Watch the next status-interval tick — you will see:
#    <'/path/to/powerline.sh right' not ready>
#    and terminal content pushed up

# 4. Confirm the fetch is slow
time bash ~/.tmux/plugins/tmux-powerline/powerline.sh right
# worst case: ~12 s
```

---

This is a focused split of #513 (which also addressed a separate emoji-width issue in `weather.sh` — that is tracked in a companion PR).

### DIFF
--- a/segments/wan_ip.sh
+++ b/segments/wan_ip.sh
@@ -38,7 +38,7 @@ run_segment() {
 			fi
 		fi
 		# Atomically acquire the lock; bail out if another invocation beat us to it
-		if ( set -o noclobber; > "$lock_file" ) 2>/dev/null; then
+		if ( set -o noclobber; : > "$lock_file" ) 2>/dev/null; then
 			(
 				exec >/dev/null 2>&1
 				trap 'rm -f "$lock_file"' EXIT
--- a/segments/wan_ip.sh
+++ b/segments/wan_ip.sh
@@ -1,5 +1,7 @@
 # shellcheck shell=bash
 # Prints the WAN IP address. The result is cached and updated according to $update_period.
+#
+# Network fetch is done in a background process so tmux rendering is never blocked.
 TMUX_POWERLINE_SEG_WAN_IP_SYMBOL="${TMUX_POWERLINE_SEG_WAN_IP_SYMBOL:-ⓦ }"
 TMUX_POWERLINE_SEG_WAN_IP_SYMBOL_COLOUR="${TMUX_POWERLINE_SEG_WAN_IP_SYMBOL_COLOUR:-255}"
 
@@ -15,34 +17,37 @@ EORC
 
 run_segment() {
 	local tmp_file="${TMUX_POWERLINE_DIR_TEMPORARY}/wan_ip.txt"
+	local lock_file="${TMUX_POWERLINE_DIR_TEMPORARY}/wan_ip_refresh.lock"
+	local update_period=900
 	local wan_ip
 
+	# Always return cached data immediately, even if stale
 	if [ -f "$tmp_file" ]; then
-		if tp_shell_is_macos || tp_shell_is_bsd; then
-			stat >/dev/null 2>&1 && is_gnu_stat=false || is_gnu_stat=true
-			if [ "$is_gnu_stat" == "true" ]; then
-				last_update=$(stat -c "%Y" "${tmp_file}")
-			else
-				last_update=$(stat -f "%m" "${tmp_file}")
-			fi
-		elif tp_shell_is_linux || [ -z "$is_gnu_stat" ]; then
-			last_update=$(stat -c "%Y" "${tmp_file}")
-		fi
-
-		time_now=$(date +%s)
-		update_period=900
-		up_to_date=$(echo "(${time_now}-${last_update}) < ${update_period}" | bc)
-
-		if [ "$up_to_date" -eq 1 ]; then
-			wan_ip=$(cat "${tmp_file}")
-		fi
+		wan_ip=$(cat "$tmp_file")
 	fi
 
-	if [ -z "$wan_ip" ]; then
-		if wan_ip=$(curl --max-time 2 -s https://whatismyip.akamai.com/); then
-			echo "${wan_ip}" >"$tmp_file"
-		elif [ -f "${tmp_file}" ]; then
-			wan_ip=$(cat "$tmp_file")
+	# Check if cache is stale or missing; if so, refresh in background
+	if ! __wan_ip_cache_is_fresh "$tmp_file" "$update_period"; then
+		# Stale-lock check: if lock is older than the maximum possible fetch time, treat it as abandoned
+		if [ -f "$lock_file" ]; then
+			local lock_mtime lock_age=0
+			lock_mtime=$(stat -c "%Y" "$lock_file" 2>/dev/null || stat -f "%m" "$lock_file" 2>/dev/null)
+			[ -n "$lock_mtime" ] && lock_age=$(( $(date +%s) - lock_mtime ))
+			if [ "$lock_age" -gt 10 ]; then
+				rm -f "$lock_file"
+			fi
+		fi
+		# Atomically acquire the lock; bail out if another invocation beat us to it
+		if ( set -o noclobber; > "$lock_file" ) 2>/dev/null; then
+			(
+				exec >/dev/null 2>&1
+				trap 'rm -f "$lock_file"' EXIT
+				local fresh_ip
+				if fresh_ip=$(curl --max-time 2 -s https://whatismyip.akamai.com/); then
+					echo "${fresh_ip}" > "$tmp_file"
+				fi
+			) &
+			disown
 		fi
 	fi
 
@@ -51,4 +56,23 @@ run_segment() {
 	fi
 
 	return 0
+}
+
+__wan_ip_cache_is_fresh() {
+	local tmp_file="$1"
+	local update_period="$2"
+	[ -f "$tmp_file" ] || return 1
+	local last_update time_now
+	if tp_shell_is_macos || tp_shell_is_bsd; then
+		stat >/dev/null 2>&1 && is_gnu_stat=false || is_gnu_stat=true
+		if [ "$is_gnu_stat" == "true" ]; then
+			last_update=$(stat -c "%Y" "$tmp_file")
+		else
+			last_update=$(stat -f "%m" "$tmp_file")
+		fi
+	else
+		last_update=$(stat -c "%Y" "$tmp_file")
+	fi
+	time_now=$(date +%s)
+	[ "$(echo "(${time_now}-${last_update}) < ${update_period}" | bc)" -eq 1 ]
 }
--- a/segments/weather.sh
+++ b/segments/weather.sh
@@ -1,6 +1,8 @@
 # shellcheck shell=bash
 # Prints the current weather in Celsius, Fahrenheits or lord Kelvins. The forecast is cached and updated with a period.
 # To configure your location, set TMUX_POWERLINE_SEG_WEATHER_(LAT|LON) in the tmux-powerline config file.
+#
+# Network fetches are done in a background process so tmux rendering is never blocked.
 
 # shellcheck source=lib/util.sh
 source "${TMUX_POWERLINE_DIR_LIB}/util.sh"
@@ -16,7 +18,6 @@ TMUX_POWERLINE_SEG_WEATHER_LON_DEFAULT="auto"
 TMUX_POWERLINE_SEG_WEATHER_CACHE_FILE_WEATHER="${TMUX_POWERLINE_DIR_TEMPORARY}/weather_cache_data.txt"
 # Add: global cache file for auto-detected location (lat/lon)
 TMUX_POWERLINE_SEG_WEATHER_CACHE_FILE_LOCATION="${TMUX_POWERLINE_DIR_TEMPORARY}/weather_cache_location.txt"
-
 
 
 generate_segmentrc() {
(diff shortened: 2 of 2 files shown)


segments/weather.sh
```diff
@@ -1,24 +1,23 @@
...
+#
+# Network fetches are done in a background process so tmux rendering is never blocked.
 
 # shellcheck source=lib/util.sh
 source "${TMUX_POWERLINE_DIR_LIB}/util.sh"
 
 TMUX_POWERLINE_SEG_WEATHER_DATA_PROVIDER_DEFAULT="yrno"
 TMUX_POWERLINE_SEG_WEATHER_UNIT_DEFAULT="c"
 TMUX_POWERLINE_SEG_WEATHER_UPDATE_PERIOD_DEFAULT="600"
-TMUX_POWERLINE_SEG_WEATHER_LOCATION_UPDATE_PERIOD_DEFAULT="86400" # 24 hours
```
erikw, line 11: Let's keep this comment

segments/weather.sh
```diff
@@ -1,24 +1,23 @@
...
+TMUX_POWERLINE_SEG_WEATHER_LOCATION_UPDATE_PERIOD_DEFAULT="86400"
 TMUX_POWERLINE_SEG_WEATHER_LAT_DEFAULT="auto"
 TMUX_POWERLINE_SEG_WEATHER_LON_DEFAULT="auto"
 
-# Global cache file for weather data
```
erikw, line 15: Let's keep this comment.

segments/weather.sh
```diff
@@ -1,24 +1,23 @@
...
 TMUX_POWERLINE_SEG_WEATHER_CACHE_FILE_WEATHER="${TMUX_POWERLINE_DIR_TEMPORARY}/weather_cache_data.txt"
-# Add: global cache file for auto-detected location (lat/lon)
```
erikw, line 17: Let's keep this comment.

segments/weather.sh
```diff
@@ -41,29 +40,69 @@ EORC
 
 run_segment() {
 	local weather=""
+	local update_period="${TMUX_POWERLINE_SEG_WEATHER_UPDATE_PERIOD:-$TMUX_POWERLINE_SEG_WEATHER_UPDATE_PERIOD_DEFAULT}"
```
erikw: The pattern is that the default value is used in `__process_settings()` if user hasn't configured a custom option. Let's follow this pattern

segments/weather.sh
```diff
@@ -88,9 +127,7 @@ __process_settings() {
 }
 
 
-# An implementation of a weather provider, just need to echo the result, run_segment() will take care of the rest
```
erikw, line 91: Let's keep this comment.

segments/weather.sh
```diff
@@ -102,14 +139,11 @@ __yrno() {
 
 	local degree=""
 
-	# There's a chance that you will get rate limited or both location APIs are not working
```
erikw, line 105: Let's keep this comment.

segments/weather.sh
```diff
@@ -102,14 +139,11 @@ __yrno() {
...
-	# Then long and lat will be "null", as literal string
 	if [ -z "$TMUX_POWERLINE_SEG_WEATHER_LAT" ] || [ -z "$TMUX_POWERLINE_SEG_WEATHER_LON" ]; then
 		tp_err_seg "Err: Unable to auto-detect your location"
 		return 1
 	fi
 
-	# Ref: https://api.met.no/doc/TermsOfService
```
erikw, line 112: Let's keep this comment.

segments/weather.sh
```diff
@@ -135,27 +169,24 @@ __yrno() {
 		degree=$(__degree_c2f "$degree")
 	fi
 	# condition_symbol=$(__get_yrno_condition_symbol "$condition" "$sunrise" "$sunset")
+	local condition_symbol
 	condition_symbol=$(__get_yrno_condition_symbol "$condition")
-	# Write the <content@date>, separated by a @ character, so we can fetch it later on without having to call 'stat'
```
erikw, line 139: Let's keep this comment.

segments/weather.sh
```diff
@@ -135,27 +169,24 @@ __yrno() {
...
 	echo "${condition_symbol} ${degree}°$(echo "$TMUX_POWERLINE_SEG_WEATHER_UNIT" | tr '[:lower:]' '[:upper:]')"
 }
 
 
-# Convert from Celcius to Lord Kelvins.
```
erikw, line 144: Let's keep this comment.

segments/weather.sh
```diff
@@ -135,27 +169,24 @@ __yrno() {
...
 __degree_c2k() {
 	local c="$1"
 	echo "${c} + 273.15" | bc
 }
 
 
-# Convert from Celcius to Fahrenheits.
```
erikw, line 151: Let's keep this comment.

segments/weather.sh
```diff
@@ -135,27 +169,24 @@ __yrno() {
...
 __degree_c2f() {
 	local c="$1"
 	echo "${c} * 9 / 5 + 32" | bc
 }
 
 
-# Get symbol for condition. Available symbol names: https://api.met.no/weatherapi/weathericon/2.0/documentation#List_of_symbols
```
erikw, line 158: Let's keep this comment.

segments/weather.sh
```diff
@@ -229,19 +260,16 @@ __read_file_split() {
 }
 
 
-# Default to empty/blank
```
erikw, line 232: Let's keep this comment.

segments/weather.sh
```diff
@@ -229,19 +260,16 @@ __read_file_split() {
...
 __read_file_content() {
 	__read_file_split "$1" 0 ""
 }
 
 
-# Default to 0
```
erikw, line 238: Let's keep this comment.

segments/weather.sh
```diff
@@ -229,19 +260,16 @@ __read_file_split() {
...
 __read_file_last_update() {
 	__read_file_split "$1" 1 0
 }
 
 
-# Read cached content if still fresh; otherwise output empty
```
erikw, line 244: Let's keep this comment.

segments/weather.sh
```diff
@@ -259,7 +287,6 @@ __weather_cache_read() {
 }
 
 
-# Write <content@timestamp> to a file, overwriting existing content
```
erikw, line 262: Let's keep this comment.

segments/weather.sh
```diff
@@ -270,76 +297,70 @@ __write_to_file_with_last_updated() {
 }
```
erikw, line 324: Let's keep this comment.

segments/weather.sh
```diff
@@ -270,76 +297,70 @@ __write_to_file_with_last_updated() {
...
 
 
-# Weather-specific cache write: only write when content is non-empty
 __weather_cache_write() {
 	local content="$1"
 	if [ -n "$content" ]; then
 		__write_to_file_with_last_updated "$TMUX_POWERLINE_SEG_WEATHER_CACHE_FILE_WEATHER" "$content"
 	else
-		# Overwrite the cache file with an error message and updated timestamp to avoid repeated fetch attempts while the provider is unavailable
```
erikw, line 279: Let's keep this comment.

segments/weather.sh
```diff
@@ -270,76 +297,70 @@ __write_to_file_with_last_updated() {
...
 		tp_err_seg "Err: Failed to fetch weather data, caching error message to avoid repeated fetch attempts"
 		__write_to_file_with_last_updated "$TMUX_POWERLINE_SEG_WEATHER_CACHE_FILE_WEATHER" "failed to fetch weather data"
 	fi
 }
 
 
-# Try setting TMUX_POWERLINE_SEG_WEATHER_LAT & TMUX_POWERLINE_SEG_WEATHER_LON automatically with GeoIP services.
```
erikw, line 286: Let's keep this comment.

segments/weather.sh
```diff
@@ -270,76 +297,70 @@ __write_to_file_with_last_updated() {
...
 __get_auto_location() {
-    local max_cache_age=$TMUX_POWERLINE_SEG_WEATHER_LOCATION_UPDATE_PERIOD
```
erikw, line 288: This looks like fixing indentation of this block. Make sure to add back the comments that were removed.

erikw (changes requested):
Interesting to see how the complexity of the weather segments grows and grows. Well, age of LLM, let's go.

segments/weather.sh
```diff
@@ -41,29 +40,69 @@ EORC
...
 			;;
 		esac
 
-		# Cache weather data if we got something.
 		__weather_cache_write "$weather"
-	fi
-
-	echo "$weather"
+	) &
+	disown
```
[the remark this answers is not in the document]
erikw: @turtle11311 please address this

segments/wan_ip.sh
```diff
@@ -15,34 +17,34 @@ EORC
...
+			(
+				exec 2>/dev/null
+				trap 'rm -f "$lock_file"' EXIT
+				touch "$lock_file"
+				local fresh_ip
+				if fresh_ip=$(curl --max-time 2 -s https://whatismyip.akamai.com/); then
+					echo "${fresh_ip}" > "$tmp_file"
+				fi
+			) &
+			disown
```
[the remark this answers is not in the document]
erikw: @turtle11311 please address this

segments/wan_ip.sh
```diff
@@ -15,34 +17,34 @@ EORC
...
 		fi
```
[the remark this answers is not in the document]
erikw: @turtle11311 please address this

segments/weather.sh
```diff
@@ -42,32 +43,76 @@ EORC
...
 }
 
 
-__process_settings() {
+__process_basic_settings() {
```
erikw, line 116: No let's keep the old name for codebase consistency.

erikw (changes requested):

segments/weather.sh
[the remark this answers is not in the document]
turtle11311: Fixed. Changed `exec 2>/dev/null` to `exec >/dev/null 2>&1` inside the background subshell in both `weather.sh` and `wan_ip.sh`. Redirecting stdout causes the subshell to close its inherited write end of the command-substitution pipe immediately, so `$(run_segment)` in `lib/powerline.sh` returns without waiting for the background `curl` to finish.

segments/weather.sh
turtle11311: Understood — happy to drop the new name. The reason we introduced `__process_basic_settings` was that `__process_settings` calls `__get_auto_location`, which makes blocking network requests. If we call the full `__process_settings` in the foreground `run_segment` path (needed to resolve `$TMUX_POWERLINE_SEG_WEATHER_UPDATE_PERIOD` before the cache-freshness check), we'd reintroduce blocking on the first call.

A couple of approaches that keep the old name:
1. Add a parameter to `__process_settings` (e.g. `skip_location=true`) so the foreground path can call it without the location lookup.
2. Inline only the `update_period` default in `run_segment` with the same `if [ -z ... ]; then export` pattern, and leave `__process_settings` unchanged for the background subshell.

Which do you prefer?

segments/wan_ip.sh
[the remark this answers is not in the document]
turtle11311: Fixed. Replaced the `[ ! -f "$lock_file" ]` check + `touch` inside the subshell with an atomic acquire outside it:

```bash
( set -o noclobber; > "$lock_file" ) 2>/dev/null || return
```

Both invocations racing to create the lock file will atomically fail except one, so only a single background `curl` is ever spawned per refresh cycle. The same fix was applied to `segments/weather.sh`.

turtle11311:
Addressing the batch of **"Let's keep this comment"** and **indentation** feedback all at once:

All comments that were inadvertently removed in the original commit have been restored — including inline comments on `__yrno`, `__degree_c2k/f`, `__get_yrno_condition_symbol`, `__read_file_content/last_update`, `__weather_cache_read`, `__write_to_file_with_last_updated`, `__weather_cache_write`, `__get_auto_location`, and the file-level variable declarations (`# 24 hours`, `# Global cache file for weather data`, etc.).

The indentation in `__get_auto_location` has also been left as-is (matching the original 4-space style), with all removed comments restored.

segments/wan_ip.sh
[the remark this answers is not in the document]
turtle11311: Same fix as the TOCTOU comment above — changed `exec 2>/dev/null` to `exec >/dev/null 2>&1` inside the background subshell, closing the inherited stdout pipe so `$(run_segment)` returns immediately.

segments/weather.sh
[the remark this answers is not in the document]
turtle11311: Same fix as `wan_ip.sh` — changed `exec 2>/dev/null` to `exec >/dev/null 2>&1` inside the background subshell, closing the inherited stdout pipe so `$(run_segment)` returns immediately.

PR #516, opened 2026-06-01
fix(paths): make sure temp directory is created

farzadmf:
## What?
Add `-p` to mkdir to make sure a non-existent parent directory doesn't break things.

## Why?
After updating to the latest version, when I started tmux after a restart (clean /tmp directory), tmux-powerline stopped working, and this change seems to have solved the issue.


erikw:
Oh good call. Thanks!

PR #518, opened 2026-08-15
feat(google_cloud): add new segment showing the Google Cloud context properties.

sciencesakura:
This new segment `google_cloud.sh` displays properties such as the account and project in your current Google Cloud environment.

## Requirement

- [gcloud CLI](https://cloud.google.com/cli)

## Screenshots

Displays the current project ID of the Google Cloud Platform:

tmux-powerline_gcloud1

When the `gcloud` command was not found:

tmux-powerline_gcloud2

Tested on Ubuntu 24.04 and macOS 26.5.


xx4h (changes requested):
Hi @sciencesakura, thanks for contributing :ok_hand:

Nice segment, i'd like to have it in. A few things before we merge:

<hr>

**1. `gcloud info` runs on every refresh**

Our default `status-interval` is `1` (`config/defaults.sh:12`), so this runs once per second, per attached client.
Measured here with gcloud 565.0.0:

```
gcloud info          1.08s wall / 0.99s CPU
gcloud config list   1.06s wall / 0.97s CPU
```

So the call might not even finish before the next refresh is due, and it keeps a core busy permanently for a value
that changes not really that often. It's the python startup that costs, not the subcommand, so moving to
`gcloud config list` wouldn't buy us anything.

I'd rather read the config directly. gcloud resolves every one of the three properties the same way. env var first,
config file second. So it's not split across two places:

| property | env override | otherwise |
| --- | --- | --- |
| `active_config_name` | `CLOUDSDK_ACTIVE_CONFIG_NAME` | content of `$CFG_DIR/active_config`, or `default` if that file doesn't exist |
| `account` | `CLOUDSDK_CORE_ACCOUNT` | `account` from the `[core]` section of `$CFG_DIR/configurations/config_<active_config_name>` |
| `project` | `CLOUDSDK_CORE_PROJECT` | `project` from that same `[core]` section |

`$CFG_DIR` being `${CLOUDSDK_CONFIG:-$HOME/.config/gcloud}`. Note the active config name decides *which*
config file the other two are read from. `GOOGLE_CLOUD_PROJECT` is not consulted by gcloud at all, i tested that.

A config file looks like this, nothing exotic:

```ini
[core]
account = staging-acct@example.com
project = staging-proj

[compute]
region = europe-west1
```

So in the segment it comes down to roughly (please test this again, i might not have thought of all the edge cases):

```bash
__google_cloud_ini_get() {
	local section="$1" key="$2" file="$3"
	[ -f "$file" ] || return 0
	awk -F' *= *' -v sect="$section" -v key="$key" '
		/^\[/ { in_sect = ($0 == "[" sect "]"); next }
		in_sect && $1 == key { print $2; exit }
	' "$file"
}

cfg_dir="${CLOUDSDK_CONFIG:-$HOME/.config/gcloud}"
config_name="${CLOUDSDK_ACTIVE_CONFIG_NAME:-$(cat "$cfg_dir/active_config" 2>/dev/null)}"
config_name="${config_name:-default}"
config_file="$cfg_dir/configurations/config_${config_name}"

account="${CLOUDSDK_CORE_ACCOUNT:-$(__google_cloud_ini_get core account "$config_file")}"
project="${CLOUDSDK_CORE_PROJECT:-$(__google_cloud_ini_get core project "$config_file")}"
```

I ran this against gcloud on the same config and it gives identical output for all four cases (plain, and each of the
three env vars set), at ~7ms instead of 1.08s.

For the awk: checked with gawk, mawk, busybox awk and one-true-awk (the one macOS ships) — same result everywhere, also for `key=value` without spaces and with extra spaces around the `=`. A multi-char `-F` being an
ERE is POSIX, so no macOS surprise there.

And it does follow changes live — gcloud is a one-shot process, so the config file *is* the state, it has nowhere
else to keep it between invocations. Verified that every mutation is on disk right after the command returns:

```
gcloud config set project proj-bbb        -> config_default:  project = proj-bbb
gcloud config configurations activate prod -> active_config:  prod
gcloud config unset project               -> project line removed again
```

If you'd rather keep the CLI, then it at least needs caching via `tp_is_tmp_valid` plus an `_UPDATE_INTERVAL` setting,
like `weather.sh` and `claude_code.sh` do — but that goes stale right after a `gcloud config set project`, which is
exactly the moment you look at the segment. What do you think?

<hr>

**2. Segment returns 1 when there's nothing to show**

```bash
[ -n "$status_text" ] && echo "${TMUX_POWERLINE_SEG_GOOGLE_CLOUD_SYMBOL}${status_text}"
```

That's the last statement, so with no project set `run_segment` exits 1, and `lib/powerline.sh:122` then kills the
whole status line in debug mode. Reproduced with a config that only has an account set:

```
rc=1 out=[]
```

Needs an explicit `return 0` at the end.

<hr>

**3. `gcloud CLI was not found` in the status bar**

I'd like to keep this in sync with the other segments, and we're mostly quiet there: `kubernetes_context.sh`, `github_notifications.sh` and `claude_code.sh` just `return 0`, `weather.sh` does `tp_err_seg` + `return 1`. Only
three older ones (`rainbarf`, `gcalcli`, `dropbox_status`) write into the bar. CONTRIBUTING describes it the same way:

> If the segment failed to execute in a fatal way return a non-zero exit code so the user can pick up the error and fix it
when debug mode is on (e.g. missing program that is needed for the segment).

Could you switch this and the invalid `DISPLAY_MODE` message over to `tp_err_seg` + `return 1`, like `weather.sh`?
The error log is the right place for it, and it doesn't take over the bar with a string the user can't get rid of.

<hr>

**4. `DISPLAY_MODE` combinations**

Right now we can pick only one of four, but with three properties there are more useful combinations. `active_config_name` + `project` is what i'd want when working with named configs, and `account_project` already
shows that combining makes sense.

Nothing of this is released yet, so we can drop the enum and take a list instead:

```
TMUX_POWERLINE_SEG_GOOGLE_CLOUD_DISPLAY_MODE="active_config_name,project"
```

With the values from **1.** already in variables, it comes down to a loop over the tokens that skips the empty ones
and joins the rest with the separator:

```bash
for prop in ${TMUX_POWERLINE_SEG_GOOGLE_CLOUD_DISPLAY_MODE//,/ }; do
	case "$prop" in
	account) value="$account" ;;
	project) value="$project" ;;
	active_config_name) value="$config_name" ;;
	*)
		tp_err_seg "Err: unknown property '$prop'"
		return 1
		;;
	esac
	[ -n "$value" ] && parts+=("$value")
done
```

Skipping the empty ones matters, otherwise you end up with a dangling separator whenever only one of the
requested properties is set. `account_project` then goes away, and `_ACCOUNT_PROJECT_SEPARATOR` becomes
just `_SEPARATOR` (same as in `kubernetes_context.sh`).

Small stuff:

* we don't need the executable bit, segments are sourced by `lib/powerline.sh`, not executed (the newer ones are 644)
* the trailing space in `SYMBOL_DEFAULT` — `kubernetes_context.sh` keeps that in the format string instead, so
overriding the symbol doesn't break the spacing

sciencesakura:
@xx4h 

Thank you for your helpful comments and review!

I'll reconsider the matter and repush the PR.

sciencesakura:
@xx4h Thanks a lot for the thorough review. I pushed the rework, here's what changed.

### 1. Reading the config directly

The segment no longer needs the `gcloud` at all.

I went through Google's docs and behavior to confirm your description before implementing it, and the solutions for all three properties were correct.

Instead of a segment-local function, I put the INI parser into `lib/`. Looking at the gcloud source, the config files are read with Python's standard [`configparser`](https://docs.python.org/3/library/configparser.html), so what we're parsing is a Python-style INI file, not a gcloud-specific format. That felt like enough to be worth having as a general utility rather than something owned by one segment, which is why it went into `lib/`. I want your opinion on this.

Tested the parser against mawk, gawk, busybox awk, and macOS awk. It worked on all of them.

| implementation     | per refresh, 100 runs |
|--------------------|-----------------------|
| mawk 1.3.4         | 5.6ms                 |
| gawk 5.3.2         | 9.1ms                 |
| busybox 1.37.0     | 7.8ms                 |
| macOS awk 20200816 | 6.9ms                 |

### 2. Returning 0 at the end of `run_segment` explicitly

Done.

### 3. Using `tp_err_seg` + `return 1`

Done.

### 4. `DISPLAY_MODE` combinations

Switched to a comma-separated list and renamed the env variable since it's no longer a "mode":

```shell
TMUX_POWERLINE_SEG_GOOGLE_CLOUD_PROPERTIES_TO_DISPLAY="active_config_name,project"
```

So `account_project` is gone, and the separator is also renamed to just `_SEPARATOR`. And there's no dangling separator regardless of the property order; `account,project` and `project,account` both give just the `account` when that's the only property set.

### Small stuff

#### 1. file permissions

I removed the executable bit from `segments/google_cloud.sh`.

#### 2. trailing space in `SYMBOL_DEFAULT`

I moved the trailing space of `SYMBOL_DEFAULT` into the format string. Now overriding the symbol no longer breaks the spacing.

PR #520, opened 2026-09-20
feat(tennis): add cached free-tier score snapshots

bensynapse:
I run Live Tennis API.

Adds an optional `tennis` segment for score snapshots, with a local player/team filter, player-major set scores, available points and snapshot age. It follows the existing segment/config generator conventions and uses curl and jq.

Only the free live-match endpoint is used. Every local instance shares a persistent 900-second attempt floor, including HTTP and transport failures (at most 96 attempts/day). Fetches run in the background; failures retain the last snapshot with its age. The segment reads one page of up to 100 matches and marks additional pages without fetching them. Setup, dedicated-key requirements and lock recovery are documented.

Validation:
- Project Super-Linter v8.3.1 full BASH check: passed.
- `bash tests/tennis.sh`: 22 scenarios pass on Bash 3.2.57 and Bash 5.2, including concurrent refreshes, restart persistence, failed calls, null scores, tiebreak points and escaped text.
- Real curl against a local HTTP fixture server and an attached tmux 3.4 terminal: background refresh, authentication header, HTTP 429 cooldown, config generation and literal rendering verified.

Live-key requests and macOS were not tested.


segments/tennis.sh
```diff
@@ -0,0 +1,112 @@
...
+	if [[ "$key" == *$'\r'* || "$key" == *$'\n'* ]]; then
+		printf '%s\n' 'tennis: the API key must be a single line' >&2
+		return 1
+	fi
+	if ! command -v curl >/dev/null || ! command -v jq >/dev/null; then
+		printf '%s\n' 'tennis: curl and jq are required' >&2
+		return 1
+	fi
+
+	# Unlike the normal temporary directory, this survives tmux and OS restarts.
```
erikw: Yes, though it would be better to be consistent with how mother segments are catching on the bolatilencache destinations. It will be confusing if some segments be a persist cache and others not. Let’s do like the other segment like weather segment

erikw (changes requested):

bensynapse:
Updated in `8277f69`: all tennis cache files now live under `${TMUX_POWERLINE_DIR_TEMPORARY}/tennis`, alongside weather. When the temporary cache is missing, the segment starts a full 15-minute wait before fetching; this keeps rebooting or clearing the cache from bypassing the free-tier request limit. The wait is shown in the segment and documented.

I also fixed duplicated match-tiebreak scores (for example, `6-4 4-6 [10-5]`). All 25 regression scenarios pass on Bash 3.2 and 5.2, the full project Super-Linter check passes, and real curl/local HTTP plus an attached tmux session verified the cache and cooldown behavior. No live-key or macOS test was run.

